PR #11, opened 2016-07-05
Support macOS

nicksnyder:


LayoutKit/UIKitSupport.swift
```diff
@@ -0,0 +1,17 @@
...
+
+import UIKit
+
+public typealias View = UIView
+
+public typealias EdgeInsets = UIEdgeInsets
+
+public typealias UserInterfaceLayoutDirection = UIUserInterfaceLayoutDirection
+
+public typealias Application = UIApplication
```
staguer: minor: often we require newlines at the end of a file, do you want to follow that in this codebase as well?

LayoutKitTests/DensityAssertions.swift
```diff
@@ -41,7 +42,12 @@ func AssertEqualDensity(actual: CGFloat, _ expected: [CGFloat: CGFloat], file: S
 
 /// Returns the expectation for the current density.
 private func expectationForCurrentDensity<T>(expected: [CGFloat: T], file: StaticString, line: UInt) -> T? {
-    let scale = UIScreen.mainScreen().scale
+    #if os(iOS)
+        let scale = UIScreen.mainScreen().scale
+    #else
+        let scale: CGFloat = 2.0
```
staguer: Looks like there's a macOS alternative as well: http://stackoverflow.com/questions/11067066/mac-os-x-best-way-to-do-runtime-check-for-retina-display

staguer:
I'm done with comments, thanks!

LayoutKit/UIKitSupport.swift
nicksnyder: Yeah, I wish there were a way to make this a warning. 

"Missing Newline At End Of File" setting doesn't work. Presumably because it is a GCC flag:
`GCC_WARN_ABOUT_MISSING_NEWLINE = YES;`

LayoutKitTests/DensityAssertions.swift
nicksnyder: thanks! fixed

PR #12, opened 2016-07-05
support animation between two layouts

nicksnyder:


LayoutKit/LayoutArrangement.swift
```diff
@@ -82,21 +116,46 @@ public struct LayoutArrangement {
...
+    private func makeSubviews(from recycler: ViewRecycler, prepareAnimation: Bool) -> [UIView] {
         let subviews = sublayouts.flatMap({ (sublayout: LayoutArrangement) -> [UIView] in
-            return sublayout.makeSubviews()
+            return sublayout.makeSubviews(from: recycler, prepareAnimation: prepareAnimation)
         })
-        if let view = layout.makeView() {
-            view.frame = frame
+        // If we are preparing an animation, then we don't want to update frames or configure views.
+        if let view = layout.makeView(from: recycler, configure: !prepareAnimation) {
+            if !prepareAnimation {
```
staguer: Looks like there are conceptually two tree walks here, one to create or update the sub/superview tree and another to set the frames and other configured properties. Sometimes we want to run them together and sometimes separately. Perhaps this logic would be cleaner as separate functions instead of passing down the flag throughout?

LayoutKit/ViewRecycler.swift
```diff
@@ -0,0 +1,78 @@
...
+        if view.tag == 0 {
+            untaggedViews.remove(view)
+        } else {
+            viewsByTag[view.tag] = nil
+        }
+    }
+
+    /// Creates or recycles a view of the desired type and tag.
+    public func makeView<T: UIView>(tag tag: Int) -> T {
+        guard let view = viewsByTag[tag] as? T else {
```
staguer: nitpick: I wouldn't use guard here since both cases are equally likely and valid, up to you though

LayoutKit/ViewRecycler.swift
```diff
@@ -0,0 +1,78 @@
...
+            t.tag = tag
+            return t
+        }
+        viewsByTag[tag] = nil
+        return view
+    }
+
+    /// Removes all unrecycled views from the view hierarchy.
+    public func purgeViews() {
+        for (_, view) in viewsByTag {
```
staguer: You can use viewsByTag.values instead.

LayoutKit/LayoutArrangement.swift
nicksnyder: It is certainly possible to do this in two passes, but I think the code would be more complicated due to view collapsing. It is nice that we can do it incrementally, e.g.:

```
if !prepareAnimation {
    for subview in subviews {
        subview.frame.offsetInPlace(dx: frame.origin.x, dy: frame.origin.y)
    }
}
```

docs/animations.md
```diff
@@ -0,0 +1,115 @@
+# Animations
+
+LayoutKit supports animating between two layouts.
+
+## Requirements
+
+1. Use the `tag` parameter to identify layouts involved in the animation.
```
nicksnyder: rename `tag` to `id`

LayoutKit/Layout.swift
```diff
@@ -82,6 +86,12 @@ public protocol Layout {
      TODO: figure out how to assert if inflexible layouts are compressed.
      */
     var flexibility: Flexibility { get }
+
+    /**
+     An identifier for the layout.
+     It is used to identify which views should be reused when animating from one layout to another.
```
staguer: I'd add a bit more detail here, talking about how applying two layouts reuses views by this id. Maybe even rename it to viewReuseId?

Also somewhere we should describe what happens if there is a duplicate.

LayoutKit/LayoutArrangement.swift
```diff
@@ -82,21 +116,46 @@ public struct LayoutArrangement {
...
+extension View {
+
+    /**
+     Similar to `addSubview()` except if `maintainCoordinates` is true, then the view's frame
+     will be adjusted so that its absolute position on the screen does not change.
+     */
+    private func addSubview(view: View, maintainCoordinates: Bool) {
+        if maintainCoordinates {
+            let frame = view.convertToAbsoluteCoordinates(view.frame)
+            addSubview(view)
```
nicksnyder: only add subview is superview is not different?

LayoutKit/Layout.swift
nicksnyder: viewId

nicksnyder:
Going to merge this into master. This is probably not the final API since we want to do #24 before releasing 2.0.

PR #16, opened 2016-07-14
Add a maxWidth parameter to LabelLayout

lucabartoletti:
maxWidth limits the width of the Label.
The idea is that in scenario like:

StackHorizontal[Label1(max 100), Label2]

Label1 can use horizontal space up to 100 Label 2 can use all the available space. All the other rules of StackLayout apply like before.

If there are other ways to achieve with current architecture please let me know. Otherwise happy to add more tests around this.

### DIFF
--- a/LayoutKit/Layouts/LabelLayout.swift
+++ b/LayoutKit/Layouts/LabelLayout.swift
@@ -25,6 +25,7 @@ public class LabelLayout: Layout {
     public let alignment: Alignment
     public let flexibility: Flexibility
     public let config: (UILabel -> Void)?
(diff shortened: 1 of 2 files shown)


nicksnyder:
Thanks for the pull request!

I think this is a specific instance of a problem that we can solve more generally. It would be useful to constrain any layout to have a minimum and/or maximum width and/or height.

I would rather solve the more general problem instead of making a special case for LabelLayout so I am going to close this PR. I have created issue #18 to discuss a solution to the more general problem. Would love to hear your thoughts there.

PR #28, opened 2016-07-21
rename `makeViews(inView view: UIView)` to `makeViews(in view: UIView)`

nicksnyder:


plivesey:
Looks good to me.

staguer:
Looks good to me too, thanks.

Issue #15, opened 2016-07-09
ReloadableViewLayoutAdapter animatable CRUD operations


afrederick1:
Hey guys, just started using the library today and have enjoyed working with it. I'm a little stuck and was hoping you could shed some light. I've got my layouts created using a UICollectionView with the `ReloadableViewLayoutAdapter`. The behavior I'm looking for is to load an initial dataset and render the layouts, then be able to insert, delete, move or update items using the collectionview animations. I can't find anything in the docs that describes animating data, and when I try to it winds up doing a reload instead. When the `update` function is called within the adapter the `incremental` value is always true, so it calls `reloadDataSync` on the collectionview instead of proceeding to call `insert`.

After doing the initial load how should I proceed to insert more items?

nicksnyder:
As you have observed, `reload` is designed to reload the entire data set. The logic you are requesting is something that we would like to support, but hasn't been implemented yet (we punted on implementing it because we haven't had that use case yet).

To support what you are asking for we need to add methods `insert`, `update`, and `delete` with signatures similar to `reload` but with appropriate modifications.

@staguer This task was originally assigned to you; do you want to work on this?

staguer:
This is indeed necessary and interesting functionality but it’s not in my list of tasks till August at the earliest. Avoiding race conditions in the async case should be fun. If anyone wants to take this up in the meantime, please feel free.

nicksnyder:
@afrederick1 @staguer 

I am not sure if it makes more sense to provide an API that requires the full layout (much simpler code and API), or an API that only asks for the incremental layout changes (more complicated code and API). Implementing the latter is tricky due to how indexes are handled in batch updates.

Here is an example of the former:

```swift
/// Basically the same as `reload` except instead of calling `reloadData` it calls the appropriate methods to trigger animations
/// Note that layoutProvider produces the entire layout, not just the parts that changed.
public func performBatchUpdates(batchUpdates: BatchUpdates,
                                layoutProvider: Void -> T,
                                width: CGFloat? = nil,
                                height: CGFloat? = nil,
                                completion: ((Void) -> Void)? = nil) {
    // implementation
}

public struct BatchUpdates {
    public let insertItemsAtIndexPaths: [NSIndexPath]
    public let deleteItemsAtIndexPaths: [NSIndexPath]
    public let moveItemsAtIndexPaths: [IndexPathMove]

    public let insertSections: NSIndexSet
    public let deleteSections: NSIndexSet
[...]

PR #29, opened 2016-07-21
Support animated batch updates on ReloadableViewLayoutAdapter (#15)

nicksnyder:


LayoutKit/Views/BatchUpdates.swift
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+
+import Foundation
+
+
+/// A set of updates to apply to a `ReloadableView`.
+public struct BatchUpdates {
+    public var insertItemsAtIndexPaths = [NSIndexPath]()
```
plivesey: I feel like the naming here isn't very swifty. It seems strange to include the type in the name of an ivar.

Is it not descriptive enough to say `insertions`, `deletes`, `moves`. Or `insertedItems`, `deletedItems`, `movedItems`?

LayoutKit/Views/BatchUpdates.swift
```diff
@@ -0,0 +1,69 @@
...
+}
+
+public func ==(lhs: SectionMove, rhs: SectionMove) -> Bool {
+    return lhs.from == rhs.from && lhs.to == rhs.to
+}
+
+/// Returns a hash value for the two integers
+/// by concatenating the lowest 32 bits of each integer.
+private func intHashValue(left: Int, _ right: Int) -> Int {
+    let leftLowBits = left % (1 << 32)
```
plivesey: a simpler solution:

```
return left.hashValue &+ right.hashValue
```

It seems like this would be appropriately evenly distributed assuming the `hashValue` of `Int` is evenly distributed. And simpler :D

LayoutKit/Views/BatchUpdates.swift
nicksnyder: Yeah, I can drop the `AtIndexPaths`. I do need to differentiate between items and sections.

```
public var insertItems = [NSIndexPath]()
public var insertSections = NSMutableIndexSet()
```

LayoutKit/Views/ReloadableView.swift
```diff
@@ -80,12 +57,28 @@ extension UICollectionView: ReloadableView {
         registerClass(UICollectionReusableView.self, forSupplementaryViewOfKind: UICollectionElementKindSectionFooter, withReuseIdentifier: reuseIdentifier)
     }
 
-    public func insert(indexPaths indexPaths: [NSIndexPath]) {
-        insertItemsAtIndexPaths(indexPaths)
-    }
+    public func perform(batchUpdates batchUpdates: BatchUpdates) {
```
plivesey: If this is public, do we want to add docs for this? Also, we should document somewhere the order of these operations because it matters

LayoutKit/Views/BatchUpdates.swift
nicksnyder: The hash function that you propose would result in guaranteed hash collisions when multiple sections are involved (e.g. [section: 0, item: x].hash == [section: 1, item: x-1].hash === [section: 2 item: x-2].hash ....) so it seems like doing doing a small amount of extra work for a better hash is prudent.

That said, it turns out that I don't end up depending on Equatable or Hashable so I could just remove this. The reason I added it to begin with was to store `Set<NSIndexPath>` but that turned out to not be necessary.

Unless you object, I think I will just leave it here in case it becomes helpful/necessary in the future.

plivesey:
```
return left.hashValue ^ right.hashValue
```
seems to be more commonly used for hashvalues

LayoutKit/Views/BatchUpdates.swift
nicksnyder: Alternatively `itemInserts`, `sectionInserts`, `itemDeletes`, `sectionDeletes`

Do you prefer past tense?
`insertedItems` vs `insertItems`

LayoutKit/Views/BatchUpdates.swift
```diff
@@ -0,0 +1,53 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+
+import Foundation
+
+
+/// A set of updates to apply to a `ReloadableView`.
+public struct BatchUpdates {
+    public var insertItems = [NSIndexPath]()
```
plivesey: ugh, i know this is annoying, but now i read this I think `insertedItems` is better. If you think this is better though, feel free to leave it.

LayoutKit/Views/BatchUpdates.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    public init(from: Int, to: Int) {
+        self.from = from
+        self.to = to
+    }
+}
+
+public func ==(lhs: SectionMove, rhs: SectionMove) -> Bool {
+    return lhs.from == rhs.from && lhs.to == rhs.to
+}
```
plivesey: new line at end of file?

LayoutKit/Views/ReloadableView.swift
```diff
@@ -80,12 +62,28 @@ extension UICollectionView: ReloadableView {
...
+            if batchUpdates.insertSections.count > 0 {
+                self.insertSections(batchUpdates.insertSections)
+            }
+            if batchUpdates.deleteSections.count > 0 {
+                self.deleteSections(batchUpdates.deleteSections)
+            }
+            for move in batchUpdates.moveSections {
+                self.moveSection(move.from, toSection: move.to)
+            }
+        }, completion: nil)
```
plivesey: is it annoying to not get a completion for this method?

LayoutKit/Views/ReloadableViewLayoutAdapter.swift
```diff
@@ -44,230 +44,129 @@ public class ReloadableViewLayoutAdapter: NSObject {
...
-        backgroundLayoutQueue.cancelAllOperations()
-        currentArrangement = arrangement
-        reloadableView?.reloadDataSync()
-    }
-
     private func reloadSynchronously<T: CollectionType, U: CollectionType where U.Generator.Element == Layout, T.Generator.Element == Section<U>>(
-        layoutFunc layoutFunc: Layout -> LayoutArrangement,
-                   layoutProvider: Void -> T,
-                   completion: (Void -> Void)? = nil) {
+        with layoutProvider: Void -> T,
```
plivesey: is with the right word here? It seems to add no additional meaning?

`reloadSynchronously(with: provider, layoutFunc: func...)`
vs
`reloadSynchronously(provider, layoutFunc: func...)`
vs
`reloadSynchronously(layoutProvider: provider, layoutFunc: func...)`

It seems like 1 and 2 are mostly equivalent. 3 gives more info and is probably what I'd go for?

LayoutKit/Views/ReloadableViewLayoutAdapter.swift
```diff
@@ -44,230 +44,129 @@ public class ReloadableViewLayoutAdapter: NSObject {
...
         let end = CFAbsoluteTimeGetCurrent()
         logger?("user: \((end-start).ms)")
         completion?()
     }
 
     private func reloadAsynchronously<T: CollectionType, U: CollectionType where U.Generator.Element == Layout, T.Generator.Element == Section<U>>(
-        layoutFunc layoutFunc: Layout -> LayoutArrangement,
-                   layoutProvider: Void -> T,
-                   completion: (Void -> Void)? = nil) {
+        with layoutProvider: Void -> T,
```
plivesey: same here

LayoutKit/Views/ReloadableViewUpdateManager.swift
```diff
@@ -0,0 +1,138 @@
...
+protocol ReloadableViewUpdateManagerDelegate: class {
+    weak var reloadableView: ReloadableView? { get }
+    var currentArrangement: [Section<[LayoutArrangement]>] { get set }
+}
+
+/// Manages updating a `ReloadableView` and its backing data.
+/// Updates may be applied incrementally, or in a batch.
+class ReloadableViewUpdateManager {
+    private weak var delegate: ReloadableViewUpdateManagerDelegate?
+    private weak var operation: NSOperation?
```
plivesey: why does this need to be weak? Add a comment

LayoutKit/Views/ReloadableViewUpdateManager.swift
```diff
@@ -0,0 +1,138 @@
...
+            BatchUpdateManager(delegate: delegate, operation: operation)
+    }
+
+    private init(delegate: ReloadableViewUpdateManagerDelegate, operation: NSOperation) {
+        self.delegate = delegate
+        self.operation = operation
+    }
+
+    func apply(partialArrangement arrangement: [Section<[LayoutArrangement]>], insertedIndexPath: NSIndexPath) {
+        // implemented by subclass
```
plivesey: Is this an abstract class? Potentially you should `fatalError` here?

LayoutKit/Views/ReloadableViewUpdateManager.swift
```diff
@@ -0,0 +1,138 @@
...
+/// Updates the `ReloadableView` incrementally as items are inserted.
+private class IncrementalUpdateManager: ReloadableViewUpdateManager {
+
+    private var pendingInsertedIndexPaths = [NSIndexPath]()
+
+    override func apply(partialArrangement arrangement: [Section<[LayoutArrangement]>], insertedIndexPath: NSIndexPath) {
+        updateReloadableView(waitUntilFinished: false) { (reloadableView: ReloadableView) in
+            self.pendingInsertedIndexPaths.append(insertedIndexPath)
+
+            // Don't modify the data while the view is moving.
```
plivesey: Is this true of all reloads when moving? Wondering if this library should handle this or leave it to the user. Because a lot of features require things to edit immediatley?

LayoutKitTests/ReloadableViewLayoutAdapterTestCase.swift
```diff
@@ -341,157 +451,11 @@ protocol TestableReloadableView: ReloadableView {
...
-            }
-        }
-    }
-
-    private func verifyVisibleItem(text text: String, frame: CGRect, line: UInt) {
-        let items = visibleCells().filter { cell -> Bool in
-            guard let label = cell.contentView.subviews.first as? UILabel else {
-                return false
-            }
-            return label.text == text
```
plivesey: skimmed over the test/sample app code

LayoutKit/Views/ReloadableView.swift
nicksnyder: added docs

LayoutKit/Views/ReloadableView.swift
nicksnyder: The completion is only for when the animations finish. From a data integrity perspective, the updates are processed synchronously.

Unfortunately, there is no completion block for UITableView batch animations, so I chose to go with a behavior that would be consistent across UICollectionView and UITableView.

LayoutKit/Views/ReloadableViewLayoutAdapter.swift
nicksnyder: Yeah I wasn't sure if this was better or not. I will change it back to option 3.

LayoutKit/Views/ReloadableViewUpdateManager.swift
nicksnyder: You are correct, the protocol shouldn't care about whether this is weak or not. Removed.

LayoutKit/Views/ReloadableViewUpdateManager.swift
nicksnyder: yeah was supposed to be an abstract class.

After taking a second look, I am going to rework this to use protocols for more clarity.

LayoutKit/Views/ReloadableViewUpdateManager.swift
nicksnyder: This incremental update manager is a user perceived performance optimization and only triggers under a very specific set of circumstances: the collection view is currently empty and there are not batch updates.

If this behavior is undesirable for certain clients for whatever reason, then this behavior can be disabled by passing an empty BatchUpdates() object. I updated the documentation to clarify.

LayoutKit/Views/BatchUpdates.swift
nicksnyder: My preference is weak, but I think present tense makes the most sense here.

This object represents a request to a reloadable view to apply a set of updates. This object is passed to the reloadable view concurrently with the new state to apply, so it doesn't feel like the updates have already happened (i.e. no past tense).

If the code was structured such that the state got updated first and then the reloadable view had to react (e.g. by handling an event callback) then past tense would make more sense.

LayoutKit/Views/BatchUpdates.swift
nicksnyder: fixed

PR #33, opened 2016-08-04
Update SizeLayout to support min/max width/height (fixes #18)

### LINKED ISSUE #18: min/max size constraints for layouts
Currently `SizeLayout` allows layouts to have a specific length along one or both dimensions.

It would be useful to be able to specify maximum width/height and minimum width/height for a nested layout.

One option would be to create a `MinSizeLayout` and a `MaxSizeLayout`. Unsure if it would make sense to combine these into one layout (e.,g. `SizeRangeLayout`), or even combine them with `SizeLayout`.

If we stick with separate layouts, then we might also want to rename `SizeLayout` to `ExactSizeLayout`

nicksnyder:
needs more tests and doc

### DIFF
--- a/LayoutKit/Layouts/SizeLayout.swift
+++ b/LayoutKit/Layouts/SizeLayout.swift
@@ -9,7 +9,7 @@
 import CoreGraphics
 
 /**
- A layout that measures to a predetermined size.
+ A layout that has size constraints.
 
  Some common use cases:
 
@@ -20,14 +20,19 @@ import CoreGraphics
  // A 1px tall divider that fills the width of its parent.
  SizeLayout<UIView>(height: 1)
  
- // A fixed height label.
- SizeLayout<UIView>(height: 50, sublayout: LabelLayout(text: "Hello"))
+ // A label with maximum width.
+ SizeLayout<UIView>(maxWidth: 100, sublayout: LabelLayout(text: "Spills onto two lines"))
+ 
+ // A label with minimum width.
+ SizeLayout<UIView>(minWidth: 100, sublayout: LabelLayout(text: "Hello", alignment: .fill))
  ```
 */
 public class SizeLayout<V: View>: BaseLayout<V>, ConfigurableLayout {
 
-    public let width: CGFloat?
-    public let height: CGFloat?
+    public let minWidth: CGFloat?
+    public let maxWidth: CGFloat?
+    public let minHeight: CGFloat?
+    public let maxHeight: CGFloat?
     public let sublayout: Layout?
 
     /**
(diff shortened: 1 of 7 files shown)


LayoutKit/Layouts/SizeLayout.swift
```diff
@@ -88,28 +151,90 @@ public class SizeLayout<V: View>: BaseLayout<V>, Layout {
...
-        // If at least one dimension is nil, then we need to measure the sublayout to inherit its value (zero if there is no sublayout).
-        if width == nil || height == nil {
-            let subsize = sublayout?.measurement(within: constrainedSize).size ?? CGSizeZero
-            if width == nil {
-                constrainedSize.width = subsize.width
-            }
-            if height == nil {
-                constrainedSize.height = subsize.height
-            }
+    public convenience init(maxSize: CGSize,
```
plivesey: Why all these convenience inits? Can't we just have one which has maxSize/minSize etc? Are you trying to enforce that only one is passed in here?

LayoutKit/Layouts/SizeLayout.swift
```diff
@@ -88,28 +151,90 @@ public class SizeLayout<V: View>: BaseLayout<V>, Layout {
...
+    }
+
+    /// Returns true if left and right are equal within a small tolerance.
+    /// A nil value is not equal to anything else.
+    private static func equals(left: CGFloat?, _ right: CGFloat?) -> Bool {
+        guard let left = left, right = right else {
+            // treat nil != nil
+            return false
         }
+        return abs(left - right) < 0.0001
```
plivesey: Create a constant for this?
Also, possibly this function belongs in a helper class? Seems like it'd probably be useful elsewhere in the project? But maybe can do that later.

LayoutKit/Layouts/SizeLayout.swift
```diff
@@ -88,28 +151,90 @@ public class SizeLayout<V: View>: BaseLayout<V>, Layout {
...
+    }
+
+    public func measurement(within maxSize: CGSize) -> LayoutMeasurement {
+        // Take the smaller of our configured max size and the given max size for measurement.
+        let availableSize = maxSize.decreasedToSize(CGSize(width: maxWidth ?? .max, height: maxHeight ?? .max))
```
plivesey: Probably ok here, but using multiple `??` for function parameters can get really slow to compile.

plivesey:
This is going to be backwards incompatible right? So major bump for the version?

Aside from comments, LGTM.

LayoutKit/Layouts/SizeLayout.swift
```diff
@@ -88,28 +151,90 @@ public class SizeLayout<V: View>: BaseLayout<V>, Layout {
...
+        let sublayoutSize = sublayoutMeasurement?.size ?? CGSizeZero
 
-        return LayoutMeasurement(layout: self, size: constrainedSize, maxSize: maxSize, sublayouts: [])
+        // Make sure that our size is in the desired range.
+        let size = sublayoutSize.increasedToSize(CGSize(width: minWidth ?? 0, height: minHeight ?? 0)).decreasedToSize(availableSize)
+
+        let sublayouts = sublayoutMeasurement.map { (measurement) in
+            return [measurement]
+        }
+        return LayoutMeasurement(layout: self, size: size, maxSize: maxSize, sublayouts: sublayouts ?? [])
```
staguer: consider doing `let sublayouts = [sublayoutMeasurement].flatMap { $0 }`, then you won't need `?? []`

LayoutKitSampleApp/UrlImageLayout.swift
```diff
@@ -16,21 +16,28 @@ import LayoutKit
...
-                    return
-                }
-                dispatch_async(dispatch_get_main_queue(), {
-                    if imageView.url == url {
-                        imageView.image = UIImage(data: data)
-                    }
-                })
-            })
-        })
+        super.init(minWidth: size.width,
```
staguer: consider adding a newline after the ( to reduce the indentation on the rest of the lines

staguer:
Looks good!

LayoutKit/Layouts/SizeLayout.swift
nicksnyder: the build script will identify the slowest functions so if it gets bad I can fix

LayoutKit/Layouts/SizeLayout.swift
nicksnyder: Will create static computed property. Can't currently create a static let because "Static stored properties not yet supported in generic types."

nicksnyder:
@plivesey Regarding backwards incompatibility, the next release will be 2.0. Doing more of a train model here than a branch model. There are already backward incompatible changes in master that haven't been released yet.

LayoutKitSampleApp/UrlImageLayout.swift
nicksnyder: will pull config out into a variable

LayoutKit/Layouts/SizeLayout.swift
nicksnyder: Actually, we might move toward making our primitive layouts unsubclassable (encourage composition over inheritance) so my point about subclassing might be irrelevant.

nicksnyder: I experimented with this approach and it just made the callsites a little more verbose, so going to leave it as it is for now.

```
SizeLayout(width: .equals(45), height: .lessThanOrEqual(50))
SizeLayout(width: 45, maxHeight: 50)
```

I promoted SizeLayout(width:height:) to a designated initializer for backwards compatibility with current subclasses.

plivesey:
Cool. LGTM then.

PR #34, opened 2016-08-04
ConfigurableLayout protocol 

staguer:
Introducing ConfigurableLayout protocol to separate instantiating a view class from configuring its properties.

### DIFF
--- a/LayoutKit.playground/Contents.swift
+++ b/LayoutKit.playground/Contents.swift
@@ -20,3 +20,5 @@ helloWorld.arrangement().makeViews()
 helloWorld.arrangement(width: 250).makeViews()
 
 helloWorld.arrangement().makeViews(direction: .RightToLeft) // just for testing; RTL happens automatically for RTL languages.
+
+CircleImagePileLayout(imageNames: [], mode: .leadingOnTop).arrangement().layout.makeView(from: ViewRecycler(rootView: nil))
--- a/LayoutKit.playground/Contents.swift
+++ b/LayoutKit.playground/Contents.swift
@@ -20,5 +20,3 @@ helloWorld.arrangement().makeViews()
 helloWorld.arrangement(width: 250).makeViews()
 
 helloWorld.arrangement().makeViews(direction: .RightToLeft) // just for testing; RTL happens automatically for RTL languages.
-
-CircleImagePileLayout(imageNames: ["earth.png", "earth.png"], mode: .trailingOnTop).arrangement().makeViews()
--- a/LayoutKit.playground/Contents.swift
+++ b/LayoutKit.playground/Contents.swift
@@ -21,4 +21,4 @@ helloWorld.arrangement(width: 250).makeViews()
 
 helloWorld.arrangement().makeViews(direction: .RightToLeft) // just for testing; RTL happens automatically for RTL languages.
 
-CircleImagePileLayout(imageNames: [], mode: .leadingOnTop).arrangement().layout.makeView(from: ViewRecycler(rootView: nil))
+CircleImagePileLayout(imageNames: ["earth.png", "earth.png"], mode: .trailingOnTop).arrangement().makeViews()
--- a/LayoutKit/ConfigurableLayout.swift
+++ b/LayoutKit/ConfigurableLayout.swift
@@ -50,7 +50,7 @@ public protocol ConfigurableLayout: Layout {
 public extension ConfigurableLayout {
     public func configure(baseTypeView baseTypeView: View) {
         guard let view = baseTypeView as? ConfigurableView else {
-            // TODO: Assert that this won't happen.
+            assertionFailure()
             return
         }
         configure(view: view)
--- a/LayoutKit/ConfigurableLayout.swift
+++ b/LayoutKit/ConfigurableLayout.swift
@@ -50,7 +50,7 @@ public protocol ConfigurableLayout: Layout {
 public extension ConfigurableLayout {
     public func configure(baseTypeView baseTypeView: View) {
         guard let view = baseTypeView as? ConfigurableView else {
-            assertionFailure()
+            assertionFailure("Expected baseTypeView \(baseTypeView) to be of type \(ConfigurableView.self) but it was of type \(baseTypeView.dynamicType)")
             return
         }
         configure(view: view)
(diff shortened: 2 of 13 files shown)


LayoutKitExampleLayouts/CircleImagePileLayout.swift
```diff
@@ -36,13 +36,17 @@ public class CircleImagePileLayout: StackLayout {
...
-        case .leadingOnTop:
-            let view: CircleImagePileView = recycler.makeView(viewReuseId: viewReuseId)
-            return view
-        case .trailingOnTop:
+    
+    override public var needsView: Bool {
+        return mode == .leadingOnTop
+    }
+    
+    public override func makeView(from recycler: ViewRecycler) -> View? {
```
nicksnyder: hmm, I don't really like how there is duplicated logic here. Also, if a layout that does this type of thing is further subclassed, then there is no type safety if a subclass does a similar thing (can't enforce that subclass's type inherits from CircleImagePileView).

I am wondering if it would be better to do one (or both) of the following:

1. Discourage/prevent subclassing of Layouts (mark them as final or turn into structs). Then layouts like this would wrap StackLayout instead of inheriting it. Perhaps even get rid of BaseLayout at the cost of some property boilerplate.

2. Pass through the type parameter for all layouts (not just SizeLayout). e.g. `public class StackLayout<V: View>: BaseLayout<V>`. This example would customize the type as `CircleImagePileView` so that it is always used. Would need to adjust the implementation of `CircleImagePileView` to have a switch between the two behaviors. This is a bit annoying because it prevents nested types: e.g. `StackLayout.Distribution -> StackLayoutDistribution`.

Thoughts?

LayoutKit/Layout.swift
```diff
@@ -74,7 +74,11 @@ public protocol Layout {
 
      MUST be run on the main thread.
      */
-    func makeView(from recycler: ViewRecycler, configure: Bool) -> View?
+    func makeView(from recycler: ViewRecycler) -> View?
+    
+    var needsView: Bool { get }
```
nicksnyder: belongs in ConfigurableLayout

LayoutKitExampleLayouts/CircleImagePileLayout.swift
nicksnyder: I thought of a third option that would perhaps be a simpler change:

Refactor this example to not use inheritance (have it own and delegate to a StackLayout instead) and then remove the parts of this PR than enabled this inheritance.

i.e. BaseLayout should not have makeView or makeConfigurableView. Rename makeConfigurableView to makeView (both aren't needed).

LayoutKit/ConfigurableLayout.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import Foundation
+
+public protocol ConfigurableLayout: Layout {
+    associatedtype ConfigurableView: View
+    func configure(view: ConfigurableView)
+}
+
+public extension ConfigurableLayout {
+    public func genericConfigure(view: View) {
```
nicksnyder: Ideas for better names
```
configure(untypedView view: View)
configure(genericView view: View)
configure(view view: View) // then the other method could be called configure(configurableView view: ConfigurableView)
```

LayoutKit/ConfigurableLayout.swift
```diff
@@ -0,0 +1,67 @@
...
+     MUST be run on the main thread.
+     */
+    func configure(view view: ConfigurableView)
+}
+
+// Implement `configure` and `makeView` from `Layout`.
+public extension ConfigurableLayout {
+    public func configure(baseTypeView baseTypeView: View) {
+        guard let view = baseTypeView as? ConfigurableView else {
+            // TODO: Assert that this won't happen.
```
nicksnyder: I am fine with an `assertionFailure` here. This represents developer error that can/should be caught in development.

LayoutKit/Layouts/LabelLayout.swift
```diff
@@ -148,19 +148,19 @@ public class LabelLayout: BaseLayout<UILabel>, Layout {
...
+        switch textType {
+        case .unattributed(let text):
+            label.text = text
+        case .attributed(let attributedText):
+            label.attributedText = attributedText
         }
-        return label
+    }
+
+    public override var needsView: Bool {
```
nicksnyder: `public override let needsView = true`?

LayoutKitExampleLayouts/CircleImagePileLayout.swift
```diff
@@ -10,15 +10,17 @@ import UIKit
 import LayoutKit
 
 /// Displays a pile of overlapping circular images.
-public class CircleImagePileLayout: StackLayout {
```
nicksnyder: might be worth adding an explanatory comment about why we are wrapping StackLayout instead of inheriting.

LayoutKit.playground/Contents.swift
```diff
@@ -20,3 +20,5 @@ helloWorld.arrangement().makeViews()
 helloWorld.arrangement(width: 250).makeViews()
 
 helloWorld.arrangement().makeViews(direction: .RightToLeft) // just for testing; RTL happens automatically for RTL languages.
+
+CircleImagePileLayout(imageNames: ["earth.png", "earth.png"], mode: .trailingOnTop).arrangement().makeViews()
```
nicksnyder: do you want to keep this in the example playground or was this just for testing?

staguer: You're right, it distracts from the Hello World introductory example.

LayoutKitExampleLayouts/CircleImagePileLayout.swift
staguer: I'll add a comment.

LayoutKit/Layouts/LabelLayout.swift
staguer: "Cannot override with a stored property 'needsView'"

LayoutKit/ConfigurableLayout.swift
```diff
@@ -0,0 +1,67 @@
...
+     MUST be run on the main thread.
+     */
+    func configure(view view: ConfigurableView)
+}
+
+// Implement `configure` and `makeView` from `Layout`.
+public extension ConfigurableLayout {
+    public func configure(baseTypeView baseTypeView: View) {
+        guard let view = baseTypeView as? ConfigurableView else {
+            assertionFailure()
```
nicksnyder: would be nice to have an error message here that contains the type of baseTypeView as well as the type of Configurable view

PR #35, opened 2016-08-19
Expose generic view parameters on all primitive layouts for easier subclassing

nicksnyder:


LayoutKit/Layouts/LabelLayout.swift
```diff
@@ -164,3 +153,18 @@ public class LabelLayout: BaseLayout<UILabel>, ConfigurableLayout {
         return true
     }
 }
+
+// MARK: - Things that belong in LabelLayout but aren't because LabelLayout is generic.
+// "Type 'TextType' nested in generic type 'LabelLayout' is not allowed"
+// "Static stored properties not yet supported in generic types"
+
+/// The types of text that a UILabel can display.
+public enum TextType {
```
plivesey: This used to be LabelLayout.TextType. Now it's just TextType. Potentially, it needs a better name now? Like `LabelTextType`?

nicksnyder: Yeah, I was wondering if I should do this or not. Will change to LabelTextType.

nicksnyder: Actually, probably LabelLayoutTextType to be consistent with StackLayoutDistribution.

PR #36, opened 2016-08-19
[Xcode] Supports Xcode 8 / Swift 2.3

KevinVitale:
This build settings allows the frameworks to be built using Xcode 8 and Swift 2.3 As a bonus, a side-effect is that Carthage will continue to work when using Xcode 8.

### DIFF
--- a/LayoutKit.xcodeproj/project.pbxproj
+++ b/LayoutKit.xcodeproj/project.pbxproj
@@ -1266,6 +1266,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 2.3;
 			};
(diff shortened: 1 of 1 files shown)


nicksnyder:
Cool, I actually haven't tested against 2.3. Is all the current code compatible with 2.3?

Will merge if tests pass.

KevinVitale:
Locally testing, two iOS tests are failing, both related to attributed string calculations:
  - `testAttributedTextDefaultFont`;  

    > _("117.5") is not equal to ("114.5") +/- ("1e-05") _

  - `testAttributedStringCustomFontPartialRange`
    > _("171.0") is not equal to ("168.0") +/- ("1e-05")_

I noticed in the latter text case, there's a conditional code path between iOS 9 & 8. For the time being, we can close this PR; I will continue to use a fork.

nicksnyder:
Thanks for the idea. I created #37 to make the tests more robust and to support Xcode 8.

PR #37, opened 2016-08-19
Support Xcode 8

nicksnyder:
Update label tests to verify that LabelLayout produces the same results as UILabel. This is actually what we care about testing and makes the tests less fragile when a new version of iOS is released (they seem to tweak the sizes of emojis and font a lot).


LayoutKit.xcodeproj/project.pbxproj
```diff
@@ -1508,6 +1510,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.linkedin.LayoutKit;
 				PRODUCT_NAME = LayoutKit;
 				SKIP_INSTALL = YES;
+				SWIFT_VERSION = 2.3;
```
plivesey: Do you want to commit this? Does this continue to work on Xcode 7?

plivesey:
LGTM, but wonder if you want this in a different branch to master

nicksnyder:
This is compatible with Xcode 7. It ignores SWIFT_VERSION.

PR #38, opened 2016-08-23
fix LabelLayout width when numberOfLines > 0

nicksnyder:


plivesey:
LGTM

staguer:
Why don't we want to use the font's Leading value? It seems desirable that we respect the value recommended / specified by the font developer.

In either case, it'd be good to add an explanation in the code saying why we do or do not want it there.

nicksnyder:
@staguer UsesFontLeading breaks the new tests cases that I added (which should pass). All we are trying to do here is replicate the behavior of UILabel, and adding UsesFontLeading causes an inconsistency in behavior.

I think the tests are all the documentation that we need. We don't need to exhaustively explain why we didn't pass in all the other available options. UsesFontLeading was the wrong option to pass in to begin with (apparently).

PR #53, opened 2016-10-13
Update InsetLayout.swift

pgherveou:
Hi there, thanks for open sourcing this cool library!
Shouldn't the config take a V here?


LayoutKit/Layouts/InsetLayout.swift
```diff
@@ -21,7 +21,7 @@ open class InsetLayout<V: View>: BaseLayout<V>, ConfigurableLayout {
                 alignment: Alignment = Alignment.fill,
                 viewReuseId: String? = nil,
                 sublayout: Layout,
-                config: ((View) -> Void)? = nil) {
+                config: ((V) -> Void)? = nil) {
```
nicksnyder: Can you change it in the convenience constructor too?

PR #57, opened 2016-10-17
layout may hold an already existing view.

mrdaios:
layout may hold an already existing view.

### DIFF
--- a/LayoutKit/ConfigurableLayout.swift
+++ b/LayoutKit/ConfigurableLayout.swift
@@ -27,6 +27,11 @@ public protocol ConfigurableLayout: Layout {
      Layouts that just position their sublayouts can return false here.
      */
     var needsView: Bool { get }
+    
+    /**
+     a View object associated to this layout.
+     */
+    var view: ConfigurableView? { get set }
 
     /**
      Configures the given view.
@@ -57,7 +62,11 @@ public extension ConfigurableLayout {
     }
 
     public func makeView(from recycler: ViewRecycler) -> View? {
-        if needsView {
+        if let view = self.view {
+            recycler.markViewAsRecycled(view)
+            return view
+        }
+        else if needsView {
             let newOrRecycledView: ConfigurableView = recycler.makeView(viewReuseId: viewReuseId)
             return newOrRecycledView
         } else {
--- a/LayoutKit/Layouts/BaseLayout.swift
+++ b/LayoutKit/Layouts/BaseLayout.swift
@@ -29,6 +29,8 @@ open class BaseLayout<V: View> {
     open var needsView: Bool {
         return viewReuseId != nil || config != nil
     }
+    
+    open var view: V?
 
     public init(alignment: Alignment, flexibility: Flexibility, viewReuseId: String? = nil, config: ((V) -> Void)?) {
         self.alignment = alignment


nicksnyder:
@mrdaios Can you explain this PR a bit more? What is the use case?

mrdaios:
@nicksnyder Similar to [StackView.ViewLayout](https://github.com/linkedin/LayoutKit/blob/master/LayoutKit/Views/StackView.swift#L116), View may be constructed by Builder(xxxBuilder.buildCustomView()), so I hope the layout associated with the existing View.

nicksnyder:
I still don't understand the point of this change. The documentation doesn't say how the view should be set or used, and there are no tests that demonstrate a use-case. If you provide some more information then I may be able to help you accomplish what you want with the current API.

The general point of LayoutKit is to define a layout which instantiates a view. StackView is a special case UIView that is similar to UIStackView, except it uses the implementation of StackLayout so layout is fast. The only reason StackView.ViewLayout exists is so that regular UIViews can be added to the StackView.

If you have a similar use-case to StackLayout then feel free to copy the technique that it uses.

mrdaios:
@nicksnyder  first tks.

there is a bug
```swift
public func makeView(from recycler: ViewRecycler) -> View? {
        if let view = self.view {
           //I missed this
            recycler.markViewAsRecycled(view)
            return view
        }
        else if needsView {
            let newOrRecycledView: ConfigurableView = recycler.makeView(viewReuseId: viewReuseId)
            return newOrRecycledView
        } else {
            return nil
        }
    }
```
I mean to be able to customize the creation of UIView, not only through the layout to create UIView
```swift

class BackgroundMiniProfileViewController: UIViewController {

    // alloc by myself
    let imageView = UIImageView()
    
    override func viewDidLoad() {
        super.viewDidLoad()
        view.backgroundColor = UIColor.white
        edgesForExtendedLayout = UIRectEdge()
        
        imageView.image = UIImage(named: "sergei")
        
        let imageLayout = SizeLayout<UIImageView>(width: 50, height: 50)
        imageLayout.view = imageView
        imageLayout.arrangement().makeViews(in: self.view)
        
        DispatchQueue.global(qos: DispatchQoS.QoSClass.userInitiated).async {
            sleep(10)
            DispatchQueue.main.async(execute: {
               //update view
                self.imageView.image = UIImage(named: "nick")
                imageLayout.arrangement().makeViews(in: self.view)
            })
        }
    }
}
```

nicksnyder:
Thanks for the code snippet. Is there a reason you want to initialize your own image view?

Here is what I would recommend:
```

class BackgroundMiniProfileViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()
        view.backgroundColor = UIColor.white
        edgesForExtendedLayout = UIRectEdge()

        let someReuseId = "someReuseId"
        let imageLayout = layoutWithImage(named: "sergei", viewReuseId: someReuseId)
        imageLayout.arrangement().makeViews(in: self.view)

        DispatchQueue.global(qos: DispatchQoS.QoSClass.userInitiated).async {
            sleep(10)
            DispatchQueue.main.async(execute: {
               //update view
                let imageLayout = layoutWithImage(named: "nick", viewReuseId: someReuseId)
        imageLayout.arrangement().makeViews(in: self.view)
            })
        }
    }

    func layoutWithImage(named: String, viewReuseId: String) -> Layout {
        return SizeLayout<UIImageView>(width: 50, height: 50, viewReuseId: viewReuseId, config: { imageView in
            imageView.image = UIImage(named: named)
        })
    }
}
```

This is a bit overkill for your situation because the view doesn't actually change size, but this is the general idea.

If you do want to instantiate your own view, then my recommendation is to create a custom Layout that takes in a UIView in its initializer and returns that in makeView (like StackView.ViewLayout).

You are correct that LayoutKit currently requires views to have a no-parameter initializer. This is something that can be improved so I created #60 to track this.

mrdaios:
@nicksnyder tks.
Is looking forward to, because it is a little cumbersome to build View.

PR #61, opened 2016-10-25
Allow deletion of arrangedSubviews from StackView

dgattey:
If an implementor wants to reuse the `StackView` with different subviews (like in a table view cell), currently, there's no way to delete the subviews, only a way to add more. This fix adds a remove method. Additionally, it fixes a documentation link.

### DIFF
--- a/LayoutKit/Views/StackView.swift
+++ b/LayoutKit/Views/StackView.swift
@@ -87,16 +87,7 @@ open class StackView: UIView {
      Deletes all subviews from the stack.
      */
     open func removeArrangedSubviews() {
-        arrangedSubviews.removeAll()
-        invalidateIntrinsicContentSize()
-        setNeedsLayout()
-    }
-
-    /**
-     Deletes all subviews from the stack.
-     */
-    open func removeArrangedSubviews() {
-        for subview in subviews {
+        for subview in arrangedSubviews {
             subview.removeFromSuperview()
         }
         arrangedSubviews.removeAll()
--- a/LayoutKit/Views/StackView.swift
+++ b/LayoutKit/Views/StackView.swift
@@ -83,6 +83,27 @@ open class StackView: UIView {
         setNeedsLayout()
     }
 
+    /**
+     Deletes all subviews from the stack.
+     */
+    open func removeArrangedSubviews() {
+        arrangedSubviews.removeAll()
+        invalidateIntrinsicContentSize()
+        setNeedsLayout()
+    }
+
+    /**
+     Deletes all subviews from the stack.
+     */
+    open func removeArrangedSubviews() {
+        for subview in subviews {
+            subview.removeFromSuperview()
+        }
+        arrangedSubviews.removeAll()
+        invalidateIntrinsicContentSize()
+        setNeedsLayout()
+    }
+
     open override func sizeThatFits(_ size: CGSize) -> CGSize {
         return stackLayout.measurement(within: size).size
     }
--- a/LayoutKitTests/StackViewTests.swift
+++ b/LayoutKitTests/StackViewTests.swift
@@ -99,35 +99,22 @@ class StackViewTests: XCTestCase {
 
     /**
      This tests to make sure we can add subviews to the stack view, lay them out, then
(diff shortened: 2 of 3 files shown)


LayoutKit/Views/StackView.swift
```diff
@@ -83,6 +83,15 @@ open class StackView: UIView {
         setNeedsLayout()
     }
 
+    /**
+     Deletes all subviews from the stack.
+     */
+    open func removeArrangedSubviews() {
+        arrangedSubviews.removeAll()
```
nicksnyder: It looks like you forgot to actually remove the views from the view hierarchy. Please add tests to StackViewTests.swift

dgattey: Check the updated PR. Thanks!

LayoutKit/Views/StackView.swift
```diff
@@ -83,6 +83,27 @@ open class StackView: UIView {
...
+        invalidateIntrinsicContentSize()
+        setNeedsLayout()
+    }
+
+    /**
+     Deletes all subviews from the stack.
+     */
+    open func removeArrangedSubviews() {
```
nicksnyder: there are now two declarations of removeArrangedSubviews

LayoutKit/Views/StackView.swift
```diff
@@ -83,6 +83,27 @@ open class StackView: UIView {
...
+        for subview in subviews {
```
nicksnyder: given the name, this should really only iterate over the views in arrangedSubviews, not all subviews.

LayoutKitTests/StackViewTests.swift
```diff
@@ -97,6 +97,51 @@ class StackViewTests: XCTestCase {
...
+        XCTAssertEqual(view3.frame, CGRect(x: 0, y: CGFloat(14+18.5), width: 33.5, height: 23))
+
+        // Now let's remove the views
+        sv.removeArrangedSubviews()
+
+        sv.frame = CGRect(origin: .zero, size: ics)
+        sv.layoutIfNeeded()
+
+        ics = sv.intrinsicContentSize
+        XCTAssertEqual(ics, CGSize.zero)
```
nicksnyder: this should probably be checked before you call sv.layoutIfNeeded on line 126?

I would also like to see an assertion that the number of subviews == 0.

LayoutKitTests/StackViewTests.swift
```diff
@@ -97,6 +97,51 @@ class StackViewTests: XCTestCase {
...
+
+        // Add them back
```
nicksnyder: Is this necessary for this test? Seems like we only need to test that the views get removed as we expect.

dgattey: I was making sure that readding didn't fail, but you make a good point that only removal should be tested here.

LayoutKitTests/StackViewTests.swift
```diff
@@ -97,6 +97,39 @@ class StackViewTests: XCTestCase {
         ])
     }
 
+    /**
+     This tests to make sure we can add subviews to the stack view, lay them out, then
+     remove them, then add them back again.
+     */
+    func testReusableStackView() {
```
nicksnyder: I think we can simplify this test further. Ideally we just want this to test removeArrangedSubviews.

```
func testRemoveArrangedSubviews() {
    let view1 = FixedSizeView(width: 7, height: 14)
    let view2 = FixedSizeView(width: 18, height: 18.5)
    let view3 = FixedSizeView(width: 33.5, height: 23)

    let stackView = StackView(axis: .vertical)

    stackView.addArrangedSubviews([view1, view2, view3])
    XCTAssertEqual(stackView.intrinsicContentSize, CGSize(width: 33.5, height: CGFloat(14 + 18.5 + 23)))
    XCTAssertEqual(stackView.subviews.count, 3)

    stackView.removeArrangedSubviews()
    XCTAssertEqual(stackView.intrinsicContentSize, CGSize.zero)
    XCTAssertEqual(stackView.subviews.count, 0)
}
```

PR #62, opened 2016-10-26
Changes needed to support ButtonLayout

nicksnyder:
3 related commits. I recommend reviewing each commit separately (in order).

### DIFF
--- a/LayoutKit/Internal/CGFloatExtension.swift
+++ b/LayoutKit/Internal/CGFloatExtension.swift
@@ -1,0 +1,65 @@
+// Copyright 2016 LinkedIn Corp.
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+
+import UIKit
+
+extension CGFloat {
+
+    /**
+     Returns the current float rounded up to the nearest fraction of a pixel
+     that the screen density supports.
+     */
+    var roundedUpToFractionalPoint: CGFloat {
+        if self == 0 {
+            return 0
+        }
+        if self < 0 {
+            return -(-self).roundedDownToFractionalPoint
+        }
+        let scale = UIScreen.main.scale
+        // The smallest precision in points (aka the number of points per hardware pixel).
+        let pointPrecision = 1.0 / scale
+        if self <= pointPrecision {
+            return pointPrecision
+        }
+        return ceil(self * scale) / scale
+    }
+
+    /**
+     Returns the current float rounded down to the nearest fraction of a pixel
+     that the screen density supports.
+     */
+    var roundedDownToFractionalPoint: CGFloat {
+        if self == 0 {
+            return 0
+        }
+        if self < 0 {
+            return -(-self).roundedUpToFractionalPoint
+        }
+        let scale = UIScreen.main.scale
+        // The smallest precision in points (aka the number of points per hardware pixel).
+        let pointPrecision = 1.0 / scale
+        if self < pointPrecision {
+            return 0
+        }
+        return floor(self * scale) / scale
+    }
+
+    /**
+     Returns the current float rounded up or down to the nearest fraction of a pixel
+     that the screen density supports.
+     */
+    var roundedToFractionalPoint: CGFloat {
+        if self == 0 {
+            return 0
+        }
+        let up = roundedUpToFractionalPoint
+        let down = roundedDownToFractionalPoint
+        return up - self <= self - down ? up : down
+    }
+}
--- a/LayoutKit/Layout.swift
+++ b/LayoutKit/Layout.swift
@@ -67,14 +67,18 @@ public protocol Layout {
     func arrangement(within rect: CGRect, measurement: LayoutMeasurement) -> LayoutArrangement
 
     /**
-     Returns a UIView for the layout or nil if the layout does not require a view (i.e. it merely positions sublayouts).
-     
-     To support animations, the implementation:
-         - SHOULD call `makeView(layoutId:)` on the provided recycler instead of instantiating the view directly.
+     Indicates whether a View object needs to be created for this layout.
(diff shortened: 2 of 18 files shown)


LayoutKit/Internal/CGFloatExtension.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     Returns the current float rounded up to the nearest fraction of a pixel
+     that the screen density supports.
+     */
+    var roundedUpToFractionalPoint: CGFloat {
+        if self <= 0 {
+            return 0
+        }
+        let scale: CGFloat = UIScreen.main.scale
```
plivesey: This doesn't seem mac safe

LayoutKit/Layout.swift
```diff
@@ -67,14 +67,17 @@ public protocol Layout {
...
+     Layouts that just position their sublayouts can return false here.
+     */
+    var needsView: Bool { get }
+    
+    /**
+     Returns a new UIView for the layout or nil if the layout does not require a view (i.e. it merely positions sublayouts).
 
      MUST be run on the main thread.
      */
-    func makeView(from recycler: ViewRecycler) -> View?
```
plivesey: So, now this won't be called if you recycle the view? Maybe worth including in the comments here.

LayoutKit/Layouts/ButtonLayout.swift
```diff
@@ -0,0 +1,223 @@
...
+ Since UIKit does not provide threadsafe methods to determine the size of a button given its content
+ ButtonLayout's implementation is not future proof for any design changes that may happen to UIButton in the future.
+ If the style of UIButton changes in the future, then ButtonLayout will need to be updated accordingly.
+
+ If future-proofing is a concern for your application, then you should not use ButtonLayout and instead implement your own
+ custom layout that uses you own custom button view (e.g. by subclassing UIControl).
+ */
+open class ButtonLayout<Button: UIButton>: BaseLayout<Button>, ConfigurableLayout {
+
+    private let type: ButtonLayoutType
```
plivesey: Why make these private? I passed them in the initializer, so seems like being able to read them may be helpful?

LayoutKit/Layouts/ButtonLayout.swift
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+    }
+}
+
+/**
+ Maps to UIButtonType.
+ This prevents LayoutKit from breaking if a new UIButtonType is added.
+ */
+public enum ButtonLayoutType {
```
plivesey: why not make this a nested enum?

ButtonLayout {
    enum Type {

LayoutKit/Internal/CGFloatExtension.swift
nicksnyder: It isn't. This file is only added to the iOS and tvOS targets.

LayoutKit/Internal/CGFloatExtension.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import UIKit
+
+extension CGFloat {
+
+    /**
+     Returns the current float rounded up to the nearest fraction of a pixel
+     that the screen density supports.
+     */
+    var roundedUpToFractionalPoint: CGFloat {
```
nicksnyder: added

nicksnyder:
I updated ButtonLayout to support images and contentEdgeInsets.

@plivesey Can you take another look at ButtonLayout.swift?

LayoutKit/Internal/CGFloatExtension.swift
```diff
@@ -0,0 +1,29 @@
...
+        if self <= 0 {
+            return 0
+        }
```
staguer: this documentation doesn't say that the result would be forced to be non-negative

LayoutKit/LayoutArrangement.swift
```diff
@@ -119,7 +119,7 @@ public struct LayoutArrangement {
             return sublayout.makeSubviews(from: recycler, prepareAnimation: prepareAnimation)
         })
         // If we are preparing an animation, then we don't want to update frames or configure views.
-        if let view = layout.makeView(from: recycler) {
+        if let view = recycler.makeOrRecycleView(havingViewReuseId: layout.viewReuseId, viewProvider: layout.makeView), layout.needsView {
```
staguer: does this mean that first the view will be made/recycled, and only after that the needsView property would be checked?

LayoutKit/Layouts/ButtonLayout.swift
```diff
@@ -0,0 +1,289 @@
...
+        return LayoutArrangement(layout: self, frame: frame, sublayouts: [])
+    }
+
+    open override func makeView() -> View {
+        return Button(type: type.buttonType)
+    }
+
+    open override func configure(view: Button) {
+        config?(view)
+        view.contentEdgeInsets = contentEdgeInsets
```
staguer: Should this support UIButton.titleEdgeInsets and UIButton.imageEdgeInsets along with UIButton.contentEdgeInsets?

LayoutKit/Layouts/ButtonLayout.swift
```diff
@@ -0,0 +1,289 @@
...
+        case .image(let image):
+            view.setImage(image, for: .normal)
+        case .size, .defaultImage:
+            break
+        }
+    }
+}
+
+/**
+ The image that appears to the left of the button's title.
```
staguer: Shouldn't we support image being on the right?
This may be out of scope of this commit but it would be good to have a suggestion for how to achieve such a layout.

LayoutKit/Layouts/ButtonLayout.swift
nicksnyder: No because `titleEdgeInsets` and `imageEdgeInsets` don't affect the actual layout. The doc for both says.

> This property is used only for positioning the title during layout. The button does not use this property to determine intrinsicContentSize and sizeThatFits(_:).

You can customize these in the config block but it doesn't change the size or position of the UIButton.

LayoutKit/LayoutArrangement.swift
nicksnyder: Good point, I will change this to make sure that `layout.needsView` is checked first.

LayoutKit/Internal/CGFloatExtension.swift
nicksnyder: Actually, in the process of trying to get attributed strings to work I made this code more general and removed this restriction. Will update the PR.

PR #65, opened 2016-11-01
update for Xcode 8.1

nicksnyder:


LayoutKitTests/ReloadableViewLayoutAdapterCollectionViewTests.swift
```diff
@@ -64,32 +64,32 @@ private class TestCollectionView: LayoutAdapterCollectionView, TestableReloadabl
         reloadDataCount += 1
     }
 
-    private override func insertItems(at indexPaths: [IndexPath]) {
```
plivesey: What's the reason for this change?

nicksnyder: Not sure why this worked in Xcode 8.0. It doesn't make sense that you could override a method and make it private :)

```
Overriding instance method must be as accessible as its enclosing type
```

PR #69, opened 2016-11-02
Make subview order deterministic

### LINKED ISSUE #66: Subviews should be in a deterministic order, regardless of view reuse
As reported by @pgherveou in #56:

> If I run twice the same layout with reuseIds set on some of my views, the order of the views changes

```
== test 2 ==

Unmanaged<AnyObject>(_value: <UIView: 0x7f92a56022c0; frame = (0 0; 350 250); layer = <CALayer: 0x610000033a20>>
   | <UIImageView: 0x7f92a5503e70; frame = (4 4; 50 50); opaque = NO; userInteractionEnabled = NO; layer = <CALayer: 0x618000234ee0>>
   | <UILabel: 0x7f92a7104440; frame = (61 4; 36.5 20.5); text = 'hello'; userInteractionEnabled = NO; layer = <_UILabelLayer: 0x60000009edc0>>
   | <UILabel: 0x7f92a5406770; frame = (58 29; 42.5 20.5); text = 'world'; userInteractionEnabled = NO; layer = <_UILabelLayer: 0x60000009dd80>>)

== test 3x ==

Unmanaged<AnyObject>(_value: <UIView: 0x7f92a56022c0; frame = (0 0; 350 250); layer = <CALayer: 0x610000033a20>>
   | <UILabel: 0x7f92a7104440; frame = (61 4; 36.5 20.5); text = 'hello'; userInteractionEnabled = NO; layer = <_UILabelLayer: 0x60000009edc0>>
   | <UILabel: 0x7f92a5406770; frame = (58 29; 42.5 20.5); text = 'world'; userInteractionEnabled = NO; layer = <_UILabelLayer: 0x60000009dd80>>
   | <UIImageView: 0x7f92a560b770; frame = (4 4; 50 50); opaque = NO; userInteractionEnabled = NO; layer = <CALayer: 0x610000036180>>)

```

This could have unexpected consequences for layering and tap handling if views overlap so LayoutKit should make sure views are in a deterministic order.

nicksnyder:
fixes #66

### DIFF
--- a/LayoutKit/LayoutArrangement.swift
+++ b/LayoutKit/LayoutArrangement.swift
@@ -146,13 +146,8 @@ extension View {
     /**
      Similar to `addSubview()` except if `maintainCoordinates` is true, then the view's frame
      will be adjusted so that its absolute position on the screen does not change.
-     
-     It doesn't do anything of view is already a subview.
      */
     fileprivate func addSubview(_ view: View, maintainCoordinates: Bool) {
-        if view.superview == self {
-            return
-        }
         if maintainCoordinates {
             let frame = view.convertToAbsoluteCoordinates(view.frame)
             addSubview(view)
--- a/LayoutKitTests/LayoutArrangementTests.swift
+++ b/LayoutKitTests/LayoutArrangementTests.swift
@@ -30,6 +30,29 @@ class LayoutArrangementTests: XCTestCase {
         XCTAssertEqual(view.subviews[1].frame, CGRect(x: 80, y: 10, width: 50, height: 50))
     }
 
+    func testSubviewOrderIsStable() {
+        // Forces the SizeLayout to produce a view.
+        let forceViewConfig: (View) -> Void = { _ in }
+        let top = SizeLayout(width: 5, height: 10, config: forceViewConfig)
+        let middle = SizeLayout(width: 5, height: 20, viewReuseId: "middle", config: forceViewConfig)
+        let bottom = SizeLayout(width: 5, height: 30, config: forceViewConfig)
+
+        let stack = StackLayout(axis: .vertical, sublayouts: [top, middle, bottom])
+        let container = View()
+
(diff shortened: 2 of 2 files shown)


LayoutKit/LayoutArrangement.swift
```diff
@@ -150,9 +150,6 @@ extension View {
      It doesn't do anything of view is already a subview.
      */
     fileprivate func addSubview(_ view: View, maintainCoordinates: Bool) {
-        if view.superview == self {
```
plivesey: Why does this fix work? What was the bug? Is it worth adding some comments to this method saying what needs to stay in order?

nicksnyder: The bug is described a bit in #66, but the tl;dr is that repeated applications of a layout to the same view hierarchy wouldn't guarantee the order of the subviews when some are reused and others are not.

The code that I removed was added as a unverified performance optimization (the internet thinks that addSubviews is slow). Unfortunately, this short circuit doesn't result in the correct ordering of the subviews (addSubview always pulls the subview to the top).

So, we can either rely on addSubviews to do the right thing (this PR) or an alternate implementation would look like this:

```swift
fileprivate func addSubview(_ view: View, maintainCoordinates: Bool) {
    if view.superview == self {
        bringSubview(toFront: view)
        return
    }
    if maintainCoordinates {
        let frame = view.convertToAbsoluteCoordinates(view.frame)
        addSubview(view)
        view.frame = view.convertFromAbsoluteCoordinates(frame)
    } else {
        addSubview(view)
    }
}
```

I am somewhat indifferent between these so I chose the "less code" option. I want to do a perf test to see if the "more code" option actually performs better.

Thoughts?

nicksnyder: Did perf test and there was no difference so going to merge as-is (less code).

PR #70, opened 2016-11-02
A non-nil viewReuseId does not force BaseLayout to have a view

### LINKED ISSUE #56: BaseLayout doesn't need a view if viewReuseId is non-nil
See @pgherveou's comment below.

nicksnyder:
fixes #56

### DIFF
--- a/LayoutKit/Layouts/BaseLayout.swift
+++ b/LayoutKit/Layouts/BaseLayout.swift
@@ -27,7 +27,7 @@ open class BaseLayout<V: View> {
     open let config: ((V) -> Void)?
 
     open var needsView: Bool {
-        return viewReuseId != nil || config != nil
+        return config != nil
     }
 
     public init(alignment: Alignment, flexibility: Flexibility, viewReuseId: String? = nil, config: ((V) -> Void)?) {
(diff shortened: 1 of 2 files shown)


nicksnyder:
Certain layouts don't require a view (e.g. StackLayout, InsetLayout) because they merely position other layouts. The exception to this is if the developer provides a config block to a StackLayout or an InsetLayout. This config block may set properties that require a view (e.g. background color), so the layout will always produce a view.

Prior to this change, setting a viewReuseId would also force a layout to produce a view. So:
```
StackLayout(axis: .vertical, sublayouts: [LabelLayout(text: "Hi")]) // produces one UILabel
StackLayout(axis: .vertical, viewReuseId: "stack", sublayouts: [LabelLayout(text: "Hi")]) // produces UILabel nested in UIView
```

After this change, these would both produce only a single UILabel. There is really no point in setting a viewReuseId on a StackLayout if it doesn't have a doesn't need a UIView so there are a few options:

1. Create a UIView anyway (this is the current behavior)
2. Don't create a UIView (this is the PR)
3. Assert/log that this doesn't really make any sense.

At least one person was surprised by (1) because they didn't realize that they didn't need to set a viewReuseId on things like StackLayout. (2) Allows developers to assign viewReuseIds without having to worry about performance or whether the layout currently needs a view. This is a bit kinder than (3) because it allows a developer to assign a viewReuseId before it may be needed.

Issue #71, opened 2016-11-10
Why there isn't update in `BatchUpdates`


RoCry:
How would I reload some rows of my table?

It seems very nature if there is updatedItems/Sections in  BatchUpdates.
What did I missed?

Thanks

nicksnyder:
Should be pretty straightforward and easy to add support for this to [BatchUpdates](https://github.com/linkedin/LayoutKit/blob/master/LayoutKit/Views/BatchUpdates.swift) and [ReloadableView](https://github.com/linkedin/LayoutKit/blob/master/LayoutKit/Views/ReloadableView.swift).

Will also need to add or update unit tests.

@RoCry Would you like to work on this and submit a pull request?

RoCry:
@nicksnyder Sure, will send the pull request soon.

nicksnyder:
fixed by #72, thanks!

PR #72, opened 2016-11-10
Add reloadItems/reloadSections in BatchUpdates and add test cases

RoCry:
As titled and #71


LayoutKitTests/ReloadableViewLayoutAdapterTestCase.swift
```diff
@@ -446,6 +451,8 @@ class ReloadableViewLayoutAdapterTestCase: XCTestCase {
         var batchUpdates = BatchUpdates()
         batchUpdates.insertItems.append(IndexPath(item: 2, section: 1))
         batchUpdates.insertSections.insert(2)
+        batchUpdates.reloadItems.append(IndexPath(item: 1, section: 0))
+        batchUpdates.reloadSections.insert(1)
```
nicksnyder: this test does both an insert and a reload on section 1, so I am worried that the reload behavior could mask a bug in the insert behavior.

To completely isolate the tests I would prefer it if you added a new section that is only used for testing section reloading.

RoCry:
@nicksnyder You are right, I tweaked the tests, should be fine now, please check it again.

nicksnyder:
Thanks!

PR #85, opened 2017-01-08
Add `isLayoutKitView` for safe view removal

inamiy:
It seems composite view's subviews e.g. `UIButton`'s subviews (`UIImageView` and `UIButtonLabel`) gets removed and displays nothing when:
	
1.`makeViews` is called twice
2. ButtonLayout has `viewReuseId` (UIButton will be reused) 

This is because `purgeViews` is removing views even though some of them are not managed by LayoutKit, so I added `isLayoutKitView` for safe view removal.

### DIFF
--- a/LayoutKitTests/ViewRecyclerTests.swift
+++ b/LayoutKitTests/ViewRecyclerTests.swift
@@ -42,7 +42,7 @@ class ViewRecyclerTests: XCTestCase {
         XCTAssertEqual(v, expectedView)
 
         recycler.purgeViews()
-        XCTAssertNil(zero.superview, "`zero` should be removed")
+        XCTAssertNil(zero.superview, "`zero` should be removed because `isLayoutKitView` is true")
     }
 
     func testNonNilIdRecycled() {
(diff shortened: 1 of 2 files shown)


nicksnyder:
Thanks for submitting, I should have a chance to look at this in a day or two.

LayoutKitTests/ViewRecyclerTests.swift
```diff
@@ -24,7 +25,24 @@ class ViewRecyclerTests: XCTestCase {
...
+
+        let recycler = ViewRecycler(rootView: root)
+        let expectedView = View()
+        let v: View? = recycler.makeOrRecycleView(havingViewReuseId: nil, viewProvider: {
+            return expectedView
+        })
+        XCTAssertEqual(v, expectedView)
+
+        recycler.purgeViews()
+        XCTAssertNil(zero.superview, "`zero` should be removed")
```
nicksnyder: "`zero` should be removed because `isLayoutKitView` is true"

inamiy: Updated in 502e6f7 (also added button test)

Issue #88, opened 2017-01-14
Animation doesn't work well for adding & removing views


inamiy:
Though animation works nicely for reused views, newly-added or removed views don't get any good visual effects since their frames are either miscalculated (when added) or not even calculated (when removed).

It would be nice to have some effects for them e.g. fade in/out, and also delaying `purgeViews` so that animation don't get interrupted.

nicksnyder:
Thanks for the report. You are correct that added views are just added and
removed views are just removed.

When I was implementing this, my thinking was that all animated views
should be in the before and after states. For add/remove this would be
implemented as opacity fade (or even translating the frame on/off screen).

I am open to making improvements, but it would be helpful if you could
provide a more concrete playground example that demonstrates the problem
you have.

Can you clarify what you mean by purgeViews interrupting animations?

On Sat, Jan 14, 2017 at 6:07 AM Yasuhiro Inami <notifications@github.com>
wrote:

Though animation works nicely for reused views, newly-added or removed
views don't get any good visual effects since their frames are either
miscalculated (when added) or not even calculated (when removed).


It would be nice to have some effects for them e.g. fade in/out, and also
delaying purgeViews so that animation don't get interrupted.




—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/linkedin/LayoutKit/issues/88>, or mute the thread

inamiy:
I added playground demos in https://github.com/linkedin/LayoutKit/pull/89.
Please take a look at "Test" page that demonstrates the awkward animation when view gets added/removed.

And, what I mean by "delay`purgeViews` so that animation don't get interrupted" means that view-removal needs to be done after fade-out animation completes.

I think LayoutKit will probably need to call `UIView.animate(...)` from library side in the future.

inamiy:
FYI I recently released https://github.com/inamiy/Zelkova (LayoutKit + Redux-like state management support).
It also includes animation demo that is simpler than https://github.com/linkedin/LayoutKit/pull/89.

staguer:
@inamiy I'm looking at Test playground page in inamiy/LayoutKit playground branch. Are you referring to the behavior of the Hello/World label that doesn't have a viewReuseId? Is the problem that it flies in from origin but you expect it to do something nicer, like fade in or out?

The way I would treat such animations is to consider the animation as being not just from some initialLayout to some finalLayout, but going through up to four distinct steps:
1) initial layout
2) pre-animation layout
[...]

PR #89, opened 2017-01-15
Improve playground & add LayoutArrangement.debugDescription

inamiy:
Added more playground pages for:

- Simple Layout Demos
- Animation (from [animations.md](https://github.com/linkedin/LayoutKit/blob/master/docs/animations.md))
- React.js/Elm-like rendering example (`Counter.xcplaygroundpage`)
- Animation failure reported in https://github.com/linkedin/LayoutKit/issues/88 (`Test.xcplaygroundpage`).

And for debugging purpose, I also added `LayoutArrangement.debugDescription`.


LayoutKit.playground/Sources/UIControl+Handler.swift
```diff
@@ -0,0 +1,14 @@
...
+
+private var controlHandlerKey: Int8 = 0
+
+extension UIControl {
+
+    public func addHandler(for controlEvents: UIControlEvents, handler: @escaping (UIControl) -> ()) {
+        let target = CocoaTarget<UIControl>(handler)
+        objc_setAssociatedObject(self, &controlHandlerKey, target, .OBJC_ASSOCIATION_RETAIN)
+
+        self.addTarget(target, action: #selector(target.sendNext), for: controlEvents)
```
staguer: Ideally this code would make sure to remove the previous target, if there was a CocoaTarget object associated with self before the new one created here, right? If so, it'd be good to have this sample code illustrate that, or at least point out the issue in a comment.

LayoutKit/LayoutArrangement.swift
```diff
@@ -141,6 +141,22 @@ public struct LayoutArrangement {
...
+
+    public var debugDescription: String {
+        return _debugDescription(0)
+    }
+
+    private func _debugDescription(_ indent: Int) -> String {
+        let t = String(repeatElement(" ", count: indent * 2))
+        let sublayouts = self.sublayouts.isEmpty
+            ? ""
+            : self.sublayouts.map { $0._debugDescription(indent + 1) }.joined(separator: "")
```
staguer: Can this be just
`let sublayouts = self.sublayouts.map { $0._debugDescription(indent + 1) }.joined()`

LayoutKit/LayoutArrangement.swift
```diff
@@ -141,6 +141,22 @@ public struct LayoutArrangement {
...
+        return"\(t)LayoutArrangement: \(frame)\n\(sublayouts)"
```
staguer: Would it be worth adding the name of the layout class here as well?

LayoutKit.playground/Sources/UIControl+Handler.swift
inamiy: Thanks for catch! Fixed in c9b57c1.

LayoutKit/LayoutArrangement.swift
inamiy: Fixed in c9b57c1.

LayoutKit/LayoutArrangement.swift
inamiy: Fixed in c9b57c1.

LayoutKit/LayoutArrangement.swift
```diff
@@ -151,8 +151,9 @@ extension LayoutArrangement: CustomDebugStringConvertible {
         let t = String(repeatElement(" ", count: indent * 2))
         let sublayouts = self.sublayouts.isEmpty
```
staguer: Do you need to check for isEmpty? I think the map/joined expression would produce the correct result in both cases and we don't need special handling for the empty case.

LayoutKit.playground/Sources/UIControl+Handler.swift
```diff
@@ -8,6 +8,7 @@ extension UIControl {
         let target = CocoaTarget<UIControl>(handler)
         objc_setAssociatedObject(self, &controlHandlerKey, target, .OBJC_ASSOCIATION_RETAIN)
 
+        self.removeTarget(target, action: #selector(target.sendNext), for: controlEvents)
```
staguer: If I'm reading the code correctly, since target is a brand new object instance, it definitely hasn't been registered as a target/action pair on the control and this call won't do anything.

I suspect that the first parameter would have to be the object returned from `objc_getAssociatedObject(self, &controlHandlerKey)` (before the setAssociatedObject call) so that we get a previous CocoaTarget instance if there was one.

LayoutKit/LayoutArrangement.swift
```diff
@@ -151,8 +151,9 @@ extension LayoutArrangement: CustomDebugStringConvertible {
...
             ? ""
-            : self.sublayouts.map { $0._debugDescription(indent + 1) }.joined(separator: "")
-        return"\(t)LayoutArrangement: \(frame)\n\(sublayouts)"
+            : self.sublayouts.map { $0._debugDescription(indent + 1) }.joined()
+        let layoutName = String(describing: layout).components(separatedBy: ".").last!
```
staguer: I know this is just debug code and there probably should always be a period in the name, it would still be a tiny bit safer to do `fullName.components(separatedBy: ".").last ?? fullName`.

One more thing, while playing with your test playground file I thought it might also help to add the viewReuseId into this output. Up to you though whether this would be adding too much text into the output thereby hurting its readability.

LayoutKit.playground/Sources/UIControl+Handler.swift
inamiy: Thanks for check! Fixed in 2a89bd6.

LayoutKit/LayoutArrangement.swift
inamiy: Fixed in 2a89bd6.

PR #112, opened 2017-03-31
Added pod badge

jingwei-huang1:


nicksnyder:
I personally think this is unnecessary and only adds clutter because the pod version should always be the same as the tagged release version. I will give @staguer access to push the pod.

jingwei-huang1:
The reason I want to add is to make sure and remind that the release and pod versions are consistent. @staguer if you think this is unnecessary, I will close this one.

Thanks for the access!

nicksnyder:
Just follow the release checklist [release checklist](https://github.com/linkedin/LayoutKit/blob/master/RELEASE-CHECKLIST.md)

PR #114, opened 2017-04-02
Fix/crash in sampleapp

jingwei-huang1:
- DWU library will replace the implementation of `cellForIndexPath` for `UITableView` and `UICollectionView`.
- Using generic of `CollectionViewController` will cause the unstable of DWU method swizzling since `CollectionViewController` will have different address because of different generic types
- `TableViewController` is working because `menuViewController` will have method swizzling for the super class at the beginning.
- To fix crash, I can think of three ways
  - Removed method swizzling (That will be hard to inject label to `LayoutKit` and implementation has to be modified.)
  - Removed generics (That's the best solution I can come with. The cons is to make the class not clean as generic one)
  - Fixing the logic in DWU (It may resolve for this time, but it might be broken again in the future if Apple changes anything on generics)

### DIFF
--- a/LayoutKitSampleApp/Benchmarks/DataBinder.swift
+++ b/LayoutKitSampleApp/Benchmarks/DataBinder.swift
@@ -6,9 +6,37 @@
 // software distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 
-import Foundation
+import UIKit
 
 protocol DataBinder {
-    associatedtype DataType
-    func setData(_ data: DataType)
+    func setData(_ data: FeedItemData)
+}
+
+enum DataBinderHelper {
+    static func setData(_ data: FeedItemData, forView view: UIView) {
+
+        if let feedItemAutoLayoutView = view as? FeedItemAutoLayoutView {
+            feedItemAutoLayoutView.setData(data)
+            return
+        }
+
+        if let feedItemManualView = view as? FeedItemManualView {
+            feedItemManualView.setData(data)
+            return
+        }
+
+        if let feedItemLayoutKitView = view as? FeedItemLayoutKitView {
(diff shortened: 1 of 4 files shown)


LayoutKitSampleApp/Benchmarks/DataBinder.swift
```diff
@@ -6,9 +6,37 @@
...
+import UIKit
 
 protocol DataBinder {
-    associatedtype DataType
-    func setData(_ data: DataType)
+    func setData(_ data: FeedItemData)
+}
+
+enum DataBinderHelper {
+    static func setData(_ data: FeedItemData, forView view: UIView) {
```
nicksnyder: it would be worth adding a comment here to explain why generics aren't used.

LayoutKitSampleApp/Benchmarks/DataBinder.swift
```diff
@@ -6,9 +6,44 @@
...
+}
+
+enum DataBinderHelper {
+
+    /// Before v4.0.2, `CollectionViewController` and `TableViewController` are using generics.
+    /// `DWURecycling` will use method swizzling to replace some `UIcollectionView` and `UITableView`
+    /// data source methods and inject label to the cells.
+    /// By using different ContentView, `TableViewController` and `CollectionViewController` will be
+    /// treated differently. That means the correct method swizzling is only done correctly for the for
+    /// time. The solution to remove generic is a fix to https://github.com/linkedin/LayoutKit/issues/103
```
staguer: This comment is unclear to me.
What does "will be treated differently" mean here specifically?
Typo in "for the for time".
It says that "`DWURecycling` will use method swizzling" -- isn't it using it currently?
It'd be good to explain the problem in more detail.
Also, a comment explaining the new approach would help too.

jingwei-huang1:
Moved to a simple solution https://github.com/linkedin/LayoutKit/pull/121

PR #117, opened 2017-04-06
Removed pre-install scripts

jingwei-huang1:
The current build script causes the error as below,
```
$ xcrun simctl delete 22FA2149-1241-469C-BF6D-462D3837DB72
Invalid device: 22FA2149-1241-469C-BF6D-462D3837DB72
The command "xcrun simctl delete 22FA2149-1241-469C-BF6D-462D3837DB72" failed and exited with 162 during .
```


jingwei-huang1:
@nicksnyder Do you remember the reason why you want to delete two simulators before running? https://github.com/linkedin/LayoutKit/pull/87

PR #118, opened 2017-04-06
Support Swift Package Manager

jingwei-huang1:
- [x] Create `Package.swift`
- [x] Move Library code to a `Sources/` folder
- [ ] Check if files in source should be flat or not
- [x] Test Unit Tests
- [ ] Create a sample project to test Swift Package Manager
- [ ] Create scripts to test Swift Package Manager in CI
- [ ] Test Cocoapods
- [ ] Test Carthage
- [ ] Update `Readme.md`

Any suggestions?

I will test more.


jingwei-huang1:
Good point. Added to checklist.

nicksnyder:
Ideally we should have a way to test the Swift package manager integration automatically (i.e. create a sample app that imports/uses LayoutKit via Swift package manager and make sure that it builds as part of CI.

PR #121, opened 2017-04-07
Fixed crash issue in sample app - Simple solution

jingwei-huang1:


LayoutKitSampleApp/Benchmarks/CollectionViewController.swift
```diff
@@ -8,6 +8,23 @@
 
 import UIKit
 
+
+/// Each view type of collection view should subclass from `CollectionViewController`
+/// If each view type uses `CollectionViewController`, it will cause the crash.
```
staguer: I'd add these two bits of text to this comment to stress what's important about these derived classes:

>Each view type of collection view should *be a non-generic* subclass from `CollectionViewController`
If each view type uses *the generic* `CollectionViewController`, it will cause the crash.

LayoutKitSampleApp/Benchmarks/CollectionViewController.swift
```diff
@@ -8,6 +8,23 @@
...
+/// The crash reproduce steps: BenchmarkViewController -> UICollectionViewUIStackFeed
+/// -> Back -> Any other UICollectionView
+/// The reason behinds this is `DWURecyclingAlert` injects benchmark and label by using
+/// method swizzling. Method swizzling doesn't work well with generics. The second time
+/// of method sizzling would not work and bring into infinite loop if different view type
+/// uses `CollectionViewController` directly.
+/// We also need to revisit this if `DWUReclingAlert` should be still used.
```
staguer: I'm not completely clear what this last line means.

nicksnyder: DWURecyclingAlert is useful because it shows how long cells take to render.

jingwei-huang1: Here is my thought, instead of using `DWURecyclingAlert`, we may take the label and benchmark part and implement method swizzling by ourselves. That will avoid some unnecessary incorrect method swizzling.

Alternatively, we may look other solutions, like other libraries, or we can inject label manually.

jingwei-huang1:
The current solution is not perfect. Do you guys think if I should create an issue to address it in the future?

LayoutKitSampleApp/Benchmarks/CollectionViewController.swift
jingwei-huang1: Thanks for the comments. Let me update it.

PR #129, opened 2017-04-14
Open ReloadableViewLayoutAdapter UIKit implementations

### LINKED ISSUE #128: Weird top scrollViewOffset on ReloadableViewLayoutAdapter with UITableView
I found a weird behaviour when using ReloadableViewLayoutAdapter with UITableView.

I am unable to remove `tableView.topInset = 35`. I think I have tried everything. Setting `automaticallyAdjustsScrollViewInsets = false` on all VCs in hierarchy and setting `tableView.sectionHeaderHeight = 0,  tableView.sectionFooterHeight = 0`. 

It is possible to set `tableView.contentInset = UIEdgeInsets(top: -35, left: 0, bottom: 0, right: 0)`  but it seems too hacky.

The only thing that fixes this is changing 
```swift
public func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
    return currentArrangement[section].header?.frame.height ?? 0
}
```
to 
```swift
public func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
    return currentArrangement[section].header?.frame.height ?? CGFloat.leastNonzeroMagnitude
}
```
in `extension ReloadableViewLayoutAdapter: UITableViewDelegate`.

Setting header height to 0 results in it using default headerHeight value. Even though I have set `tableView.sectionHeaderHeight = 0` it still adds topInset. Am I missing something?

I do not think default implementation should be `CGFloat.leastNonzeroMagnitude`. Better solution IMO would be to make method open.

Rep2:
Making ReloadableViewLayoutAdapter UIKit implementations (UITableViewDelegate, UITableViewDataSource, UICollectionViewDelegateFlowLayout, UICollectionViewDataSource) open in order to allow for easy customisation of default behaviour.

This allows for potentially breaking overrides of given methods. Unfortunately, Swift currently does not have something like `__attribute((objc_requires_super))`. That being said, I think this change is still beneficial.

Fixes #128

### DIFF
--- a/LayoutKit.xcodeproj/project.pbxproj
+++ b/LayoutKit.xcodeproj/project.pbxproj
@@ -209,6 +209,7 @@
 		44F968181E4263DC00392763 /* TextViewLayoutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44F968161E42639500392763 /* TextViewLayoutTests.swift */; };
 		44F968191E4263DC00392763 /* TextViewLayoutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44F968161E42639500392763 /* TextViewLayoutTests.swift */; };
 		44F9681A1E42640400392763 /* TextViewLayout.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44F968141E425F5D00392763 /* TextViewLayout.swift */; };
+		AD2C36441EA5AFB500550A03 /* ReloadableViewLayoutAdapterCollectionViewOverrideTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AD2C36421EA5AF9500550A03 /* ReloadableViewLayoutAdapterCollectionViewOverrideTests.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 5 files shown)


LayoutKit/Views/ReloadableViewLayoutAdapter+UICollectionView.swift
```diff
@@ -12,15 +12,15 @@ import UIKit
 
 extension ReloadableViewLayoutAdapter: UICollectionViewDelegateFlowLayout {
```
nicksnyder: Please add a doc comment to each extension that says overrides to these methods must call super.

nicksnyder: Also, can you add some test code to make sure that overriding actually works.

Rep2: Added `/// - Warning: Not calling super may result in unexpected behaviour`. Would another comment better suite your standard? I've noticed often usage of `MUST`, but it isn't one of the official field headers and so is not bolded.

nicksnyder: Thanks! I would just clarify that it refers to subclasses and not the implementation in the extension
`Warning: Subclasses that override this method must call super.`

Rep2: Done. Tests have also been added

PR #130, opened 2017-04-18
Open ReloadableView UITableView and UICollectionView implementations

### LINKED ISSUE #124: ReloadableViewLayoutAdapter batchUpdates animation style
Hi. I am finding ReloadableViewLayoutAdapter quite useful, but most of the methods are not open making it hard to use it for anything but the simplest of cases. Is there any reason for this?

For example, I need to change UITableView batchUpdates animation style. Method `public func perform(batchUpdates: BatchUpdates)` in `extension UITableView: ReloadableView` uses .automatic and cannot be overridden.

`open func reload<T: Collection, U: Collection>` in `ReloadableViewLayoutAdapter` is open but `var currentArrangement` is `internal(set)` because of which it is needed to call `super. reload`. As such it is not possible custom reload and call custom batchUpdates.

I ended up implementing custom ReloadableViewLayoutAdapter. Any better ideas? Would it be possible to make open some of the methods?

Rep2:
Making implementation of methods `func reloadDataSynchronously()` and `func perform(batchUpdates: BatchUpdates)` open in order to simplify behaviour modification.

In order to be able to override methods declared in extensions, those methods need to be exposable to objc. In order to do so, `BatchUpdates` was changed to `class` and inherited `NSObject`.

Fixes #124

### DIFF
--- a/Sources/Views/BatchUpdates.swift
+++ b/Sources/Views/BatchUpdates.swift
@@ -10,7 +10,7 @@ import Foundation
 
 
 /// A set of updates to apply to a `ReloadableView`.
-public struct BatchUpdates {
+public class BatchUpdates: NSObject {
     public var insertItems = [IndexPath]()
     public var deleteItems = [IndexPath]()
     public var reloadItems = [IndexPath]()
@@ -21,7 +21,9 @@ public struct BatchUpdates {
     public var reloadSections = IndexSet()
     public var moveSections = [SectionMove]()
 
-    public init() { }
+    public override init() {
+        super.init()
+    }
 }
 
 /// Instruction to move an item from one index path to another.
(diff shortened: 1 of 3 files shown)


Sources/Views/BatchUpdates.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 
 
 /// A set of updates to apply to a `ReloadableView`.
-public struct BatchUpdates {
+public class BatchUpdates: NSObject {
```
staguer: Please add a comment explaining why NSObject base class is useful, maybe with an example usecase.

jingwei-huang1:
@Rep2 I fixed the CI issue. Could you please pull and merge from master branch and push again?

Rep2:
@jingwei-huang1 done

PR #135, opened 2017-04-25
Fix ci builder

jingwei-huang1:
- [x] Remove iOS 9.3 simulators for tests to fix broken CI build
- [x] Investigate why the build still passes with an error
- [ ] Investigate how to enable iOS 9.3 simulator


staguer:
I asked on https://github.com/travis-ci/docs-travis-ci-com/issues/1106 about 9.3 being gone. I wonder if there's a better place for that.

Issue #133, opened 2017-04-20
LayoutKit cells not updating in TableView and CollectionView examples


domasn:
`4.0.2` release with no modifications.

Exampe of [`TableViewControllerFeedItemLayoutKitView`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/TableViewController.swift#L18) uses [`FeedItemLayoutKitView`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift#L13) as a view wrapped inside a [`TableCell`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/TableViewController.swift#L60).

On the first glance (on left) it seems to be working just fine - item with poster number `0` followed by item with poster number `1`. However after we scroll a bit item poster numbers lose their order (on right).. We see `5` followed by `0`:

![reuse bug](https://cloud.githubusercontent.com/assets/7569279/25217956/9f6f1814-25b1-11e7-9bff-4773e2bbaf8e.png)

-----

### Problem 1 - Reusing
After a bit of debugging we can see that [`layoutSubviews()` inside `FeedItemLayoutKitView`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift#L37) does not get called for reused cells leaving cells with outdated content.

This can be solved by adding  `setNeedsLayout()` at the end of [`setData(_:)` inside `FeedItemLayoutKitView`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift#L27).

### Problem 2 - Height
Even with `setNeedsLayout()` inserted, table view does not call [`intrinsicContentSize` on `FeedItemLayoutKitView`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift#L33) for every cell before display. Even though example uses `UITableViewAutomaticDimension` as row height, differnet height cells created with LayoutKit would simply not work.

-----

Please see simple [playground code with live examples in my gist](https://gist.github.com/domasn/58a42251a81e26ffda447e267b363ed2).

Thank you for working on this amazing project! 👏

nicksnyder:
Thanks for the great bug report!

[...]

PR #136, opened 2017-04-26
Fix FeedItemLayoutKitView

domasn:
Implements fixes for `FeedItemLayoutKitView` issues described in #133.

- Fix for content not updating after cell reuse:
A simple `setNeedsLayout()`
- Fix for different cell heights being ignored:
A private height constraint.

### DIFF
--- a/LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
+++ b/LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
@@ -14,6 +14,14 @@ class FeedItemLayoutKitView: UIView, DataBinder {
 
     private var layout: FeedItemLayout? = nil
 
+    private var intrinsicContentSizeCalculated = false
+    
+    private lazy var heightConstraint: NSLayoutConstraint = {
+        let constraint = NSLayoutConstraint(item: self, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: self.bounds.height)
+        constraint.isActive = true
+        return constraint
+    }()
+    
     func setData(_ data: FeedItemData) {
         let posterProfile = ProfileCardLayout(
             name: data.posterName,
@@ -26,6 +34,10 @@ class FeedItemLayoutKitView: UIView, DataBinder {
         layout = FeedItemLayout(actionText: data.actionText, posterProfile: posterProfile, posterComment: data.posterComment, contentLayout: content, actorComment: data.actorComment)
         
         setNeedsLayout()
+        
+        if intrinsicContentSizeCalculated {
+            heightConstraint.constant = sizeThatFits(CGSize(width: bounds.width, height: .greatestFiniteMagnitude)).height
+        }
     }
 
     override func sizeThatFits(_ size: CGSize) -> CGSize {
@@ -33,6 +45,7 @@ class FeedItemLayoutKitView: UIView, DataBinder {
     }
 
     override var intrinsicContentSize: CGSize {
+        intrinsicContentSizeCalculated = true
         return sizeThatFits(CGSize(width: CGFloat.greatestFiniteMagnitude, height: CGFloat.greatestFiniteMagnitude))
     }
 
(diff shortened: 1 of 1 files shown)


LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
```diff
@@ -24,13 +32,20 @@ class FeedItemLayoutKitView: UIView, DataBinder {
 
         let content = ContentLayout(title: data.contentTitle, domain: data.contentDomain)
         layout = FeedItemLayout(actionText: data.actionText, posterProfile: posterProfile, posterComment: data.posterComment, contentLayout: content, actorComment: data.actorComment)
+        
+        setNeedsLayout()
+        
+        if intrinsicContentSizeCalculated {
```
nicksnyder: do you really need to wait for intrinsicContentSize to be called or do you just want bounds.width to be non-zero?

```
if bounds.width > 0 {
    heightConstraint.constant = sizeThatFits(CGSize(width: bounds.width, height: .greatestFiniteMagnitude)).height
}

```

LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
```diff
@@ -13,7 +13,13 @@ import ExampleLayouts
 class FeedItemLayoutKitView: UIView, DataBinder {
 
     private var layout: FeedItemLayout? = nil
-
+    
+    private lazy var heightConstraint: NSLayoutConstraint = {
+        let constraint = NSLayoutConstraint(item: self, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: self.bounds.height)
```
staguer: This is a minor point, but please see point 1.10 of https://github.com/linkedin/swift-style-guide#1-code-formatting for the preferred formatting style for long lines such as this one.

LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
```diff
@@ -24,6 +30,12 @@ class FeedItemLayoutKitView: UIView, DataBinder {
 
         let content = ContentLayout(title: data.contentTitle, domain: data.contentDomain)
         layout = FeedItemLayout(actionText: data.actionText, posterProfile: posterProfile, posterComment: data.posterComment, contentLayout: content, actorComment: data.actorComment)
+        
+        setNeedsLayout()
```
nicksnyder: can you throw in a comment to explain why this is necessary

LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
```diff
@@ -24,6 +30,12 @@ class FeedItemLayoutKitView: UIView, DataBinder {
...
+        
+        if bounds.width > 0 {
```
nicksnyder: ditto, add comment

LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
domasn: This is actually the same as for [`FeedItamManualView#L123`](https://github.com/linkedin/LayoutKit/blob/master/LayoutKitSampleApp/Benchmarks/FeedItemManualView.swift#L123).

Do you think a comment like this would be sufficient?
```swift
// Assure that `layoutSubviews` is called
setNeedsLayout()
```

LayoutKitSampleApp/Benchmarks/FeedItemLayoutKitView.swift
domasn:
```swift
// Only calculate height for valid width
if bounds.width > 0 {
```
Is this clear?

domasn:
@staguer @nicksnyder This has been here for a while now... What do you think?

nicksnyder:
Thanks! Sorry for the delay.

PR #141, opened 2017-05-16
LayoutAdapter with automatic animations

Rep2:
I found this code quite useful in my project, but it might not be something you would find is in scope of LayoutKit. If so fell free to close this PR.

Currently `ReloadableViewLayoutAdapter` uses `layoutProvider: @escaping (Void) -> T` to create new Layouts every time reload is called. 

Instead of recreating all Layouts, it is possible to use `BatchUpdates` to create new Layouts and Section only when needed. This improves performance and simplifies Layout creation. Also, if `BatchUpdates` is used to animate content change, but all of the Layouts are recreated, only animated cells will have newly created Layouts which can lead to weird behaviour.

Furthermore, it is possible to create `BatchUpdates` from model being presented. Using this two functionalities it is possible to simply present model and have animations calculated and performed automatically. 

For example, all of the animations bellow are done by presenting model using `LayoutAdapterWithAutomaticBatchUpdates` without any custom calculations.

animationsexample

### DIFF
--- a/Sources/Views/BatchUpdatesFromArrayDifference.swift
+++ b/Sources/Views/BatchUpdatesFromArrayDifference.swift
@@ -1,0 +1,57 @@
+extension BatchUpdates {
+    static func calculate(old: [[Any]], new: [[Any]], elementCompareCallback: (Any, Any) -> Bool) -> BatchUpdates {
+        var deleted = [IndexPath]()
+        var inserted = [IndexPath]()
+        var moved = [ItemMove]()
+
+        for i in 0..<min(new.count, old.count) {
+            var remainingNewValues = new[i]
+                .enumerated()
+                .map { (element: $0.element, offset: $0.offset, alreadyFound: false) }
+
+            outer: for oldValue in old[i].enumerated() {
+                for newValue in remainingNewValues
+                    where elementCompareCallback(oldValue.element, newValue.element) && !newValue.alreadyFound {
+
+                        if oldValue.offset != newValue.offset {
+                            moved.append(ItemMove(from: IndexPath(row: oldValue.offset, section: i), to: IndexPath(row: newValue.offset, section: i)))
+                        }
+
+                        remainingNewValues[newValue.offset].alreadyFound = true
+
+                        continue outer
+                }
+
+                deleted.append(IndexPath(row: oldValue.offset, section: i))
+            }
+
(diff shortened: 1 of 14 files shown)


staguer:
Hi @Rep2 ,  this looks very useful and interesting! Thank you for submitting this PR. I'm still reading through it and thinking about it and I'll have more questions later. 

(Meanwhile I just have one minor comment that `class LayoutAdapterCacheHandler` is repeated in two files in this PR :-) )

Sources/Views/BatchUpdatesFromArrayDifference.swift
```diff
@@ -0,0 +1,57 @@
+extension BatchUpdates {
+    static func calculate(old: [[Any]], new: [[Any]], elementCompareCallback: (Any, Any) -> Bool) -> BatchUpdates {
```
nicksnyder: calculate is not a great name for method, old and new are not great names for variables, and `any` type should be avoided.

Sources/Views/BatchUpdatesFromArrayDifference.swift
```diff
@@ -0,0 +1,57 @@
+extension BatchUpdates {
```
nicksnyder: I don't think this needs to be a separate extension on BatchUpdates

Sources/Views/BatchUpdatesFromArrayDifference.swift
```diff
@@ -0,0 +1,57 @@
...
+                for newValue in remainingNewValues
+                    where elementCompareCallback(oldValue.element, newValue.element) && !newValue.alreadyFound {
+
+                        if oldValue.offset != newValue.offset {
+                            moved.append(ItemMove(from: IndexPath(row: oldValue.offset, section: i), to: IndexPath(row: newValue.offset, section: i)))
+                        }
+
+                        remainingNewValues[newValue.offset].alreadyFound = true
+
+                        continue outer
```
nicksnyder: don't use constructs like this, it makes the code hard to read.

Rep2:
Thanks for the response. I agree with and understand your position. I have no problem if you modify and/or use this code outside of this PR.

If you will proceed with this idea feel free to add me to the future PRs. I will respond to the comments as soon as I find the time.

PR #147, opened 2017-07-18
Version bump 4.0.2 to 4.0.3

staguer:


staguer:
Discarding this pull request because one of the changes (BatchUpdates changing from struct to class) requires major version number increment.

Issue #142, opened 2017-05-25
High memory usage if load a large amout of items at once


bcse:
I noticed the memory usage is very high if I changed items number.
I tested by changing this line: https://github.com/linkedin/LayoutKit/blob/2c89e0b63552/LayoutKitSampleApp/FeedBaseViewController.swift#L41
And here is my test results:

- 1000: 12 MB
- 2000: 130 MB*
- 3000: 300 MB*
- 4000: 500 MB*
- 5000: 800 MB* -- On some devices, this caused out-of-memory

`*` This is top memory usage, it decreased after it hit the top

I also did another test. I call reload with 1000 items at the first time. And then wait a second, call reload with 2000 items. And then wait a second, call reload with 3000 items. And keeps doing this until 10000 items. In this case, memory is around 40 MB.

nicksnyder:
I don't understand the first test scenario. FeedBaseViewController is a base class, what child class were you performing the test on?

> I also did another test.  In this case, memory is around 40 MB.

You are saying the peak memory usage in the second scenario is 40 MB?

ReloadableViewLayoutAdapter stores all layouts in memory and does a few transformations on reload (so it may have peak memory usage multiples more than necessary). We don't have lists of items that long in the LinkedIn app, so it isn't much of a concern for us.

If this is a real-world concern for you, PRs to improve memory usage are welcome :)

bcse:
For the first scenario, I am tested with FeedTableViewController.

Currently I am load all data from a local DB and feed in to ReloadableViewLayoutAdapter at once. Now I think I should try to load them incrementally. Thanks for your help. 👍

PR #150, opened 2017-08-01
Incremental update is changed to use chunking

evnik:
This should reduce memory footprint on asynchronous loading of lots of items with `ReloadableViewLayoutAdapter` (#142).
The reason for that huge memory allocation is `items` is copied every time the `append` method is [called](https://github.com/linkedin/LayoutKit/blob/bdc0fdc9407d9e4cefc773401bf51e22779201bc/Sources/Views/ReloadableViewLayoutAdapter.swift#L136), means for each item in the section. The reason of that is `append` makes a copy in case the array passed somewhere else, and it is [passed](https://github.com/linkedin/LayoutKit/blob/bdc0fdc9407d9e4cefc773401bf51e22779201bc/Sources/Views/ReloadableViewLayoutAdapter.swift#L138) to the `Section` initializer for each iteration.
For some reason, ARC doesn't release memory used for array copies quickly, but does it slowly and only after all the items are loaded.
I changed the code to apply partial arrangements using chunks (16 is used as a chunk size), so the `items` should be copied less often.
I tested the issue #142 using Simulator with items number set to `10000`. Before my changes memory footprint in peak was more than *3 GB*, after changes it is less than *150 MB*.

### DIFF
--- a/Sources/Views/ReloadableViewLayoutAdapter.swift
+++ b/Sources/Views/ReloadableViewLayoutAdapter.swift
@@ -15,6 +15,9 @@ import UIKit
  */
 open class ReloadableViewLayoutAdapter: NSObject, ReloadableViewUpdateManagerDelegate {
 
+    static let incrementalUpdateChunkSize = 16
+    static let incrementalUpdateChunkingThreshold = 4 * incrementalUpdateChunkSize
+
     let reuseIdentifier = String(describing: ReloadableViewLayoutAdapter.self)
 
     /// The current layout arrangement.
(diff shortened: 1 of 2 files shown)


Sources/Views/ReloadableViewLayoutAdapter.swift
```diff
@@ -127,19 +145,26 @@ open class ReloadableViewLayoutAdapter: NSObject, ReloadableViewUpdateManagerDel
...
-                    updateManager.apply(partialArrangement: partialArrangement, insertedIndexPath: insertedIndexPath)
+                    if (itemIndex <= ReloadableViewLayoutAdapter.incrementalUpdateChunkingThreshold
+                        || itemIndex % ReloadableViewLayoutAdapter.incrementalUpdateChunkSize == 0)
+                    {
+                        applyPartialArrangement(header: header, items: items, footer: footer, pendingArrangement: pendingArrangement, insertedIndexPaths: insertedIndexPaths, updateManager: updateManager)
+                        insertedIndexPaths.removeAll()
+                    }
+                }
+                
+                if insertedIndexPaths.isEmpty == false {
```
nicksnyder: `if !insertedIndexPaths.isEmpty`

Sources/Views/ReloadableViewLayoutAdapter.swift
```diff
@@ -118,6 +121,21 @@ open class ReloadableViewLayoutAdapter: NSObject, ReloadableViewUpdateManagerDel
             BatchUpdateManager(delegate: self, operation: operation)
 
         operation.addExecutionBlock { [weak operation] in
+
+            func applyPartialArrangement(
```
nicksnyder: A code comment that explains why this function here would be nice

Sources/Views/ReloadableViewLayoutAdapter.swift
```diff
@@ -127,19 +145,26 @@ open class ReloadableViewLayoutAdapter: NSObject, ReloadableViewUpdateManagerDel
...
 
                     items.append(layoutFunc(itemLayout))
+                    insertedIndexPaths.append(IndexPath(item: itemIndex, section: sectionIndex))
 
-                    let partialSection = Section(header: header, items: items, footer: footer)
-                    var partialArrangement = pendingArrangement
-                    partialArrangement.append(partialSection)
-                    let insertedIndexPath = IndexPath(item: itemIndex, section: sectionIndex)
-                    updateManager.apply(partialArrangement: partialArrangement, insertedIndexPath: insertedIndexPath)
+                    if (itemIndex <= ReloadableViewLayoutAdapter.incrementalUpdateChunkingThreshold
```
nicksnyder: Can you add a comment that explains why it updates for each item for the first chunk and then chunks at a time after that

PR #153, opened 2017-08-21
Allow isSelectable overridden in TextViewLayout Configure

jingwei-huang1:
Originally, we don't allow `isSelectable` overridden in `TextViewLayout` because there was a view issue. After re-tested it on iOS 10/11, it looks OK. As a result, `isSelectable` moved to before `configure`, so that it can be overridden in configure block.

### DIFF
--- a/Sources/Layouts/TextViewLayout.swift
+++ b/Sources/Layouts/TextViewLayout.swift
@@ -130,8 +130,7 @@ open class TextViewLayout<TextView: UITextView>: BaseLayout<TextView>, Configura
     // MARK: - overriden methods
 
     open override func configure(view textView: TextView) {
-        /// `isSelectable` is default to false, but it allows overriding in `configure` block
-        /// By enabling it, it may have side effects
+        // `isSelectable` is default to false, but it allows overriding in `configure` block
         textView.isSelectable = false
 
         super.configure(view: textView)
--- a/Sources/Layouts/TextViewLayout.swift
+++ b/Sources/Layouts/TextViewLayout.swift
@@ -26,7 +26,7 @@ open class TextViewLayout<TextView: UITextView>: BaseLayout<TextView>, Configura
 
     /// Don't change `textContainerInset`, `lineFragmentPadding` and `usesFontLeading` in `configure` closure that's passed to init.
     /// By changing those, it will cause the Layout's size calculation to be incorrect. So they will be reset by using parameters from initializer.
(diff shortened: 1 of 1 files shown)


Sources/Layouts/TextViewLayout.swift
```diff
@@ -130,6 +130,10 @@ open class TextViewLayout<TextView: UITextView>: BaseLayout<TextView>, Configura
     // MARK: - overriden methods
 
     open override func configure(view textView: TextView) {
+        /// `isSelectable` is default to false, but it allows overriding in `configure` block
+        /// By enabling it, it may have side effects
```
nicksnyder: Can you either document what kind of side effects or just remove `textView.isSelectable = false` entirely.

If it is safe for isSelectable to be true, then we should probably just remove this.

jingwei-huang1: I only tested it on iOS 10/11 and I haven't fully tested on iOS 8/9. Let me remove this.

Sources/Layouts/TextViewLayout.swift
```diff
@@ -130,6 +130,9 @@ open class TextViewLayout<TextView: UITextView>: BaseLayout<TextView>, Configura
     // MARK: - overriden methods
 
     open override func configure(view textView: TextView) {
+        // `isSelectable` is default to false, but it allows overriding in `configure` block
+        textView.isSelectable = false
```
nicksnyder: My point was that we should document why we are defaulting it to false (presumably because it cause some issue in iOS 8/9). If there is no longer a reason then we should remove this line entirely.

chriscuecal:
Is there any updates on if this can be merged? Thanks!

Sources/Layouts/TextViewLayout.swift
jingwei-huang1: After I retested iOS 8/9 and I couldn't find the issue anymore, so I will remove this entire line for the time being.

jingwei-huang1:
@chriscuecal Sorry for the late reply. I was trying to test on iOS 8/9 to see if I can reproduce the issue. Since it can't be reproduced anymore, I removed the entire line and chose the default value from UITextView instead. I will merge it once @nicksnyder approves it.

chriscuecal:
@jingwei-huang1 awesome thank you!

spekke:
Sweet! Thanks!

PR #155, opened 2017-08-30
Use max size for measuring overlay sublayouts

sergeimikhan:
In OverlayLayout when measure overlays and backgrounds we need to use maxSize of primary layout measurement.

### DIFF
--- a/LayoutKitTests/OverlayLayoutTests.swift
+++ b/LayoutKitTests/OverlayLayoutTests.swift
@@ -100,7 +100,7 @@ class OverlayLayoutTests: XCTestCase {
      */
     func testPrimaryLayoutSmallerThanOverlay() {
         let primaryLayout = SizeLayout<View>(minWidth: 2, minHeight: 2)
-        let sublayout = SizeLayout<View>(minWidth: 10, minHeight: 10, alignment: .fill)
+        let sublayout = SizeLayout<View>(minWidth: 10, minHeight: 10)
 
         let overlay = OverlayLayout(primary: primaryLayout, overlay: [sublayout])
         let arrangement = overlay.arrangement(width: 2.0, height: 2.0)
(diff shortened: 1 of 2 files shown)


nicksnyder:
Please describe the issue you are fixing with an example and add a unit test.

sergeimikhan:
Have added unit test.
Here is the example with wrong behavior:

```swift
let primaryLayout = SizeLayout<View>(minWidth: 25) {
    $0.backgroundColor = .blue
}
let sublayout = SizeLayout<View>(minWidth: 100, alignment: .fillTrailing) {
    $0.backgroundColor = .red
}
OverlayLayout(primary: primaryLayout, overlay: [sublayout]).arrangement(width: 200, height: 200)
```
Here is the result with current implementation:

image

But with the fix:

image

Problem is caused by flexible primary layout size.
Thanks.

sergeimikhan:
Have added test. @dgattey, please help to check. Thanks.

LayoutKitTests/OverlayLayoutTests.swift
```diff
@@ -95,6 +95,27 @@ class OverlayLayoutTests: XCTestCase {
     }
 
     /**
+     Tests primary layout smaller than overlay.
+     Sublayouts should have same measurements.
+     */
+    func testPrimaryLayoutSmallerThanOverlay() {
+        let primaryLayout = SizeLayout<View>(minWidth: 2, minHeight: 2)
+        let sublayout = SizeLayout<View>(minWidth: 10, minHeight: 10, alignment: .fill)
```
dgattey: Does this work fine without `fill` too? In no circumstance should the overlay appear bigger than the primary layout.

sergeimikhan:
Works fine without ".fill" too.

PR #158, opened 2017-09-08
Fixing async issue on completion callback for perform batch updates

e-liu:
Currently the completion callback when performing a reload with batch updates is being called prematurely because performing batch updates happens asynchronously on the animation thread. We should use the completion block provided by performBatchUpdates(:) instead

### DIFF
--- a/Sources/Views/ReloadableViewLayoutAdapter.swift
+++ b/Sources/Views/ReloadableViewLayoutAdapter.swift
@@ -92,13 +92,13 @@ open class ReloadableViewLayoutAdapter: NSObject, ReloadableViewUpdateManagerDel
             return sectionLayout.map(layoutFunc)
         }
         if let batchUpdates = batchUpdates {
-            reloadableView?.perform(batchUpdates: batchUpdates)
+            reloadableView?.perform(batchUpdates: batchUpdates, completion: completion)
         } else {
             reloadableView?.reloadDataSynchronously()
+            completion?()
         }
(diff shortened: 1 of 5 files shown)


Sources/Views/ReloadableViewLayoutAdapter.swift
```diff
@@ -92,13 +92,13 @@ open class ReloadableViewLayoutAdapter: NSObject, ReloadableViewUpdateManagerDel
...
         }
         if let batchUpdates = batchUpdates {
-            reloadableView?.perform(batchUpdates: batchUpdates)
+            reloadableView?.perform(batchUpdates: batchUpdates, completion: completion)
         } else {
             reloadableView?.reloadDataSynchronously()
+            completion?()
         }
         let end = CFAbsoluteTimeGetCurrent()
         logger?("user: \((end-start).ms)")
```
nicksnyder: this should probably also move into the completion block

nicksnyder:
You need to update the tests. They are failing.

e-liu:
Yeah just saw, updated!

e-liu:
@nicksnyder I moved the `pendingInsertedIndexPaths.removeAll()` in `flushPendingInserts` call back outside of the completion block. It seems like calling it directly after calling reload and not waiting until completion is the right thing to do, is this correct?

nicksnyder:
@e-liu yes, that makes sense to me. We handle them synchronously by queueing up the animation. We don't need to wait for the animation to finish before considering them handled.

PR #173, opened 2017-11-22
fix LabelLayout async init used UILabel().

ghmake1y:
If we initialize the LabelLayout asynchronously with an NSAttributedString, most of the time we don't have to set the font, but the default access to UILabel(). Font, this is not safe.I don't quite understand why this is designed in this way.

### DIFF
--- a/LayoutKitTests/LabelLayoutTests.swift
+++ b/LayoutKitTests/LabelLayoutTests.swift
@@ -60,6 +60,17 @@ class LabelLayoutTests: XCTestCase {
 
         XCTAssertEqual(arrangement.makeViews().frame.size, label.intrinsicContentSize)
     }
+    
+    func testAsyncAttributedLabel() {
+        DispatchQueue.global(qos: .userInitiated).async {
+            let attributedText = NSAttributedString(string: "Async", attributes: [NSAttributedStringKey.font: UIFont.helvetica(size: 42)])
+            let arrangement = LabelLayout(attributedText: attributedText).arrangement()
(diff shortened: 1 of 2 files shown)


jingwei-huang1:
@ghmake1y Thanks for your PR. Could you elaborate why `UILabel()` isn't safe? We want to match `LabelLayout` font the same as `UILabel` by default

nicksnyder:
@ghmake1y Does the test you added fail without your change?

ghmake1y:
@nicksnyder Yes,the test function named 'testAsyncAttributedLabel' is failed if i don't change default font.

ghmake1y:
@jingwei-huang1 You can see code of function testAsyncAttributedLabel in LabelLayoutTests.swit.If async call UILabel() will be crashed,it will be trigger Main Thread Checker: UI API called on a background.

siuying:
@jingwei-huang1 UILabel() is not safe because LayoutKit can call it in background queue. When use LabelLayout in ReloadableViewLayoutAdapter, I saw Xcode complaint ``Main Thread Checker: UI API called on a background thread: -[UIView init]
PID: 26099, TID: 2008031, Thread name: (none), Queue name: ReloadableViewLayoutAdapter (QOS: USER_INITIATED), QoS: 25``.

You can reproduce this issue by running the LayoutKitSampleApp and open the "FeedCollectionViewController" example.

siuying:
If using system font is not ideal, maybe we can replace it with ``UIFont.preferredFont(forTextStyle: .body)``?

PR #176, opened 2017-12-23
Add Cocoapods static_framework support

soniccat:
I see Cocoapods RC (1.4.0.rc.1) supports swift static frameworks. Added that to avoid creating a dynamic library.
The latest Cocoapods RC can be installed using this command ```sudo gem install cocoapods --pre```

### DIFF
--- a/Sources/ViewRecycler.swift
+++ b/Sources/ViewRecycler.swift
@@ -6,7 +6,7 @@
 // software distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 
-import ObjectiveC
+import Foundation
 
 /**
  Provides APIs to recycle views by id.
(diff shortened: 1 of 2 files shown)


Sources/ViewRecycler.swift
```diff
@@ -6,7 +6,7 @@
 // software distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 
-import ObjectiveC
+import Foundation
```
soniccat: Replaced because of using NSNumber here. Had compile errors without that.

staguer:
Hi @soniccat ! Can you please link me to some documentation for this `static_framework` flag?

Also, is this something that should be determined by those entities that import LayoutKit rather than by LayoutKit itself?

soniccat:
@staguer hi, here it is http://blog.cocoapods.org/CocoaPods-1.4.0/ I see they have written it recently

nicksnyder:
Lgtm based on reading doc. Any downsides?

chenxiao0228:
Enabling this will cause the binary to be linked statically.

If the pod consumer has multiple targets using LayoutKit, LayoutKit will be liked multiple times.

Agree with  @staguer it might be more appropriate for the consumers to decide how it will be linked.

Cocoapods 1.5.0 is adding support for this use case, FYI.

PR #177, opened 2018-01-04
Add animationType to BatchUpdates

fmirault:
and use it for insertRows / deleteRows / reloadRows / insertSections / deleteSections / reloadSections methods.

Possible addition to #124 for animations.

### DIFF
--- a/Sources/Views/ReloadableView.swift
+++ b/Sources/Views/ReloadableView.swift
@@ -120,13 +120,13 @@ extension UITableView: ReloadableView {
 
         // Update items.
         if batchUpdates.insertItems.count > 0 {
-            insertRows(at: batchUpdates.insertItems, with: .automatic)
+            insertRows(at: batchUpdates.insertItems, with: batchUpdates.animationType)
         }
         if batchUpdates.deleteItems.count > 0 {
-            deleteRows(at: batchUpdates.deleteItems, with: .automatic)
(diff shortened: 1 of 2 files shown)


fmirault:
Hi ! 

Still no news since january 4th for my PR, maybe I missed something or I have to ask for a review ?

Regards,

Sources/Views/ReloadableView.swift
```diff
@@ -120,27 +120,27 @@ extension UITableView: ReloadableView {
 
         // Update items.
         if batchUpdates.insertItems.count > 0 {
-            insertRows(at: batchUpdates.insertItems, with: .automatic)
+            insertRows(at: batchUpdates.insertItems, with: batchUpdates.animationType)
```
staguer: Can we control the animations on the collection view in a similar way? See line 70 in this file.

fmirault: Unfortunately, in UIKit collection view doesn't have the same animation properties, you can only insert / delete / reload / move without any more possibility. :( That's why I only updated table view animations

fmirault:
Collection View isn't working like Table View, so I don't really see how I can achieve the same effect on it (it's an UIKit limitation in my opinion).

Sources/Views/ReloadableView.swift
fmirault: More than 2 months since my last answer... Could you please at least read it ?

PR #194, opened 2018-03-12
Adding builders to the pod. Creating target for objc code.

staguer:
Adding the builders to the LayoutKitObjC pod.
This required moving the objc enums to the header files, since objc builders cannot depend on Swift implementation.
Since the cross-language integration depends on header files that match the module name, creating a new LayoutKitObjC target for Xcode build to use to match the cocoa pods build.
Verified that build succeeds for Xcode and `pod lib lint`.

### DIFF
--- a/Sources/ObjCSupport/Builders/LOKStackLayoutBuilder.h
+++ b/Sources/ObjCSupport/Builders/LOKStackLayoutBuilder.h
@@ -8,6 +8,23 @@
 
 #import "LOKBaseLayoutBuilder.h"
 
+typedef NS_ENUM(NSInteger, LOKStackLayoutDistribution) {
+    LOKStackLayoutDistributionDefault = 0,
+    LOKStackLayoutDistributionLeading = 1,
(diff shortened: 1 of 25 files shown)


Sources/ObjCSupport/Builders/LOKStackLayoutBuilder.h
```diff
@@ -8,6 +8,23 @@
 
 #import "LOKBaseLayoutBuilder.h"
 
+typedef NS_ENUM(NSInteger, LOKStackLayoutDistribution) {
+    LOKStackLayoutDistributionDefault = 0,
```
li-eweng: You don't need to define that it's = 0. This is the default behavior. Also you don't need to define any associated value because the extension just converts it to swift enum anyway right? Same comment for the other NS_ENUMs

staguer: True. TBH I just copied these out of the LayoutKitObjC-Swift.h auto-generated file, I can clean these up.

PR #199, opened 2018-04-13
Fix link for ButtonLayout

bellebethcooper:
ButtonLayout was linked to LabelLayout.swift


staguer:
@bellebethcooper Thank you for the fix!

PR #211, opened 2018-05-23
Remove required from LOKInsetLayout and add arrangeLayout method with…

li-eweng:


Sources/ObjCSupport/LOKLayoutArrangement.swift
```diff
@@ -31,6 +31,13 @@ import CoreGraphics
             height: height.isFinite ? height : nil))
     }
 
+    @objc public class func arrangeLayout(_ layout: LOKLayout, origin: CGPoint, width: CGFloat, height: CGFloat) -> LOKLayoutArrangement {
```
staguer: It seems a bit strange to use CGPoint for origin but not CGSize for size.

PR #212, opened 2018-05-30
Fixed Swift ObjC Wrapper of layout

drumnkyle:
There could be an issue if you inherit from one of the provided layouts from LayoutKit. This fix makes it so that if you do that, it will call the correct overridden methods.

### DIFF
--- a/Sources/ObjCSupport/LOKLayout.swift
+++ b/Sources/ObjCSupport/LOKLayout.swift
@@ -20,6 +20,16 @@ import CoreGraphics
 
 extension LOKLayout {
     var unwrapped: Layout {
+        /*
+         Need to check to see if `self` is one of the LayoutKit provided layouts.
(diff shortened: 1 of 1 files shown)


Sources/ObjCSupport/LOKLayout.swift
```diff
@@ -20,6 +20,24 @@ import CoreGraphics
...
+        let allLayoutClasses = [
+            LOKButtonLayout.self,
+            LOKInsetLayout.self,
+            LOKLabelLayout.self,
+            LOKOverlayLayout.self,
+            LOKSizeLayout.self,
+            LOKStackLayout.self,
+            LOKTextViewLayout.self
+        ]
+        if let object = self as? NSObject {
```
staguer: please add a comment explaining why we need to do this

PR #214, opened 2018-05-31
Remove `with` prefix from blocks & LOK base builder. build -> layout

drumnkyle:
- Removed `with` prefix from every block property on the ObjC builders
- Removed LOKBaseLayoutBuilder
- Changed `build` function to `layout` property


Sources/ObjCSupport/Builders/LOKSizeLayoutBuilder.m
```diff
@@ -10,126 +10,150 @@
...
+@property (nonatomic, nullable) NSString *privateViewReuseId;
+@property (nonatomic, nullable) Class privateViewClass;
+
+@property (nonatomic, nullable) id<LOKLayout> privateSublayout;
+@property (nonatomic) CGFloat privateMinWidth;
+@property (nonatomic) CGFloat privateMaxWidth;
+@property (nonatomic) CGFloat privateMinHeight;
+@property (nonatomic) CGFloat privateMaxHeight;
+@property (nonatomic) CGFloat privateWidth;
+@property (nonatomic) CGFloat privateHeight;
```
staguer: `privateWidth` and `privateHeight` aren't used to construct the layout, so we don't need private properties for them

Sources/ObjCSupport/Builders/LOKButtonLayoutBuilder.m
```diff
@@ -10,96 +10,119 @@
...
-- (LOKButtonLayoutBuilder * _Nonnull (^)(void(^ _Nullable)(UIButton *_Nonnull)))withConfig {
+- (LOKButtonLayoutBuilder * _Nonnull (^)(void(^ _Nullable)(UIButton *_Nonnull)))config {
     return ^LOKButtonLayoutBuilder *(void(^ _Nullable config)(UIButton *_Nonnull)){
-        self.configure = config;
+        self.privateConfigure = config;
         return self;
     };
 }
 
+- (LOKInsetLayoutBuilder * _Nonnull (^)(EdgeInsets))insets {
```
staguer: what do you think about renaming `EdgeInsets` and `View` to `LOKEdgeInsets` and `LOKView` in this PR?

Sources/ObjCSupport/Builders/LOKBaseLayoutBuilder.m
```diff
@@ -1,29 +0,0 @@
...
-#import "LOKInsetLayoutBuilder.h"
-
-@implementation LOKBaseLayoutBuilder
-
-- (id<LOKLayout>)build {
-    NSAssert(NO, @"This method is expected to be overridden by a superclass.");
-    return nil;
-}
-
-- (id<LOKLayout>)layout {
```
jingwei-huang1: Do you think if we should add a protocol to have `- (id<LOKLayout>)layout` and `BaseLayout` elements, like `alignment`, `flexibility`, `viewReuseId`, `viewClass`. So any layout builder should implement those for preventing different variants of APIs.

Sources/ObjCSupport/Builders/LOKSizeLayoutBuilder.m
drumnkyle: It's for the setter/getter and ability to just call it like a property. So there is no storage needed. Do you want me to remove it and call the actual methods? I think I should leave it as-is.

Sources/ObjCSupport/Builders/LOKButtonLayoutBuilder.m
drumnkyle: Yeah. Sure. I can do that.

Sources/ObjCSupport/Builders/LOKBaseLayoutBuilder.m
drumnkyle: Yeah. That would be good. But how can we do that, because the return value of all of these things are a specific type.

drumnkyle: Resolved in person. Even though the protocol will declare `id<LOKLayout>` and `id<LOKLayoutBuilder>` as the types, when we implement the protocol, we can use the concrete type and it will satisfy the requirement. I will make these changes now.

jingwei-huang1: Thanks 👍

PR #218, opened 2018-07-12
Making property and function as open and exposing "items" property in LOKLayoutArrangementSection

nbadakh:
Making property and function as open and exposing "items" property in LOKLayoutArrangementSection

1. Added open access specifier for following cases so we can override it in other modules:
   - Making `LOKBaseLayout.needsView` property as open
   - Making [LOKBaseLayout configure] function open
2. Creating a public property "items" in LOKLayoutArrangementSection, so we can access "items" from the object of LOKLayoutArrangementSection wherever needed in code
3. Added `horizontallyHighlyFlexible` static constant in LOKFlexibility

### DIFF
--- a/Sources/ObjCSupport/LOKFlexibility.swift
+++ b/Sources/ObjCSupport/LOKFlexibility.swift
@@ -22,6 +22,7 @@ import Foundation
     @objc public static let high = LOKFlexibility(flexibility: .high)
     @objc public static let low = LOKFlexibility(flexibility: .low)
 
+    @objc public static let horizontallyHighlyFlexible = LOKFlexibility(flexibility: Flexibility(horizontal: Flexibility.highFlex, vertical: Flexibility.inflexibleFlex))
     @objc public static let horizontallyFlexible = LOKFlexibility(flexibility: Flexibility(horizontal: Flexibility.defaultFlex, vertical: Flexibility.inflexibleFlex))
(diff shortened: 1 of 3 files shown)


Sources/ObjCSupport/LOKFlexibility.swift
```diff
@@ -22,6 +22,7 @@ import Foundation
     @objc public static let high = LOKFlexibility(flexibility: .high)
     @objc public static let low = LOKFlexibility(flexibility: .low)
 
+    @objc public static let horizontallyHighFlexible = LOKFlexibility(flexibility: Flexibility(horizontal: Flexibility.highFlex, vertical: Flexibility.inflexibleFlex))
```
staguer: horizontallyHighFlexible -> horizontallyHighlyFlexible

Sources/ObjCSupport/LOKLayoutArrangementSection.swift
```diff
@@ -10,7 +10,9 @@ import Foundation
 
 @objc open class LOKLayoutArrangementSection: NSObject {
     let unwrapped: Section<[LayoutArrangement]>
+    @objc public let items: [LOKLayoutArrangement]
```
staguer: please add similar properties for header and footer as well

Sources/ObjCSupport/LOKFlexibility.swift
nbadakh: Done

Sources/ObjCSupport/LOKLayoutArrangementSection.swift
nbadakh: Done

PR #219, opened 2018-07-13
Added direction support in `LOKLayoutArrangement` class

nbadakh:
- Added new method `makeViews(in:direction:)` in `LOKLayoutArrangement` class.
- This method takes `UserInterfaceLayoutDirection` as parameter.
- So caller can send layout direction to `LOKLayoutArrangement` similar to `LayoutArrangement`.

### DIFF
--- a/Sources/ObjCSupport/LOKLayoutArrangement.swift
+++ b/Sources/ObjCSupport/LOKLayoutArrangement.swift
@@ -38,8 +38,14 @@ import CoreGraphics
             height: height.isFinite ? height : nil))
     }
 
-    @objc public func makeViews(in view: View?) {
-        layoutArrangement.makeViews(in: view)
+    @discardableResult
+    @objc public func makeViews(in view: View?) -> View {
+        return layoutArrangement.makeViews(in: view)
+    }
+
+    @discardableResult
+    @objc public func makeViews(in view: View?, direction: UserInterfaceLayoutDirection) -> View {
+        return layoutArrangement.makeViews(in: view, direction: direction)
     }
 
     @objc public func makeViews() -> View {


Sources/ObjCSupport/LOKLayoutArrangement.swift
```diff
@@ -42,6 +42,10 @@ import CoreGraphics
         layoutArrangement.makeViews(in: view)
     }
 
+    @objc public func makeViews(in view: View?, direction: UserInterfaceLayoutDirection) {
```
staguer: The first parameter should probably be non-optional. Otherwise, what result would the caller get if they called this method and passed in nil there? There's no return value. Same issue applies to the method above this one, it'd be nice to fix it there as well while we're making breaking changes.

nbadakh: Hi Sergei,

`layoutArrangement.makesViews(in:direction:)` method creates the views for the layout and adds them as subviews to the provided view. If no view is provided, then a new one is created and returned.
So `nil` is possible input for this method.

Thank you,
\- Nitin

staguer: There's a view returned from the method that's on `LayoutArrangement` but not from the method that's being implemented here in `LOKLayoutArrangement`.

nbadakh: Now added return type to `LOKLayoutArrangement.makeViews(in:direction)` and `LOKLayoutArrangement.makeViews(in:)`
Also added `@discardableResult` to avoid warnings like "result of call makeViews(in:) is unused"

PR #221, opened 2018-07-19
Exposed few more properties in LOK classes

nbadakh:
Exposed following properties:
- LOKStackLayout.sublayouts
- LOKInsetLayout.sublayout
- LOKLabelLayout.text
- LOKLabelLayout.attributedString

### DIFF
--- a/Sources/ObjCSupport/LOKInsetLayout.swift
+++ b/Sources/ObjCSupport/LOKInsetLayout.swift
@@ -9,19 +9,30 @@
 import CoreGraphics
 
 @objc open class LOKInsetLayout: LOKBaseLayout {
+    @objc public let insets: EdgeInsets
+    @objc public let alignment: LOKAlignment
+    @objc public let viewClass: View.Type
+    @objc public let sublayout: LOKLayout
+    @objc public let configure: ((View) -> Void)?
+
     @objc public init(insets: EdgeInsets,
                       alignment: LOKAlignment? = nil,
                       viewReuseId: String? = nil,
                       viewClass: View.Type? = nil,
                       sublayout: LOKLayout,
                       configure: ((View) -> Void)? = nil) {
+        self.insets = insets
+        self.sublayout = sublayout
+        self.alignment = alignment ?? .fill
+        self.viewClass = viewClass ?? View.self
+        self.configure = configure
         let layout = InsetLayout(
-            insets: insets,
-            alignment: alignment?.alignment ?? .fill,
+            insets: self.insets,
+            alignment: self.alignment.alignment,
             viewReuseId: viewReuseId,
-            sublayout: sublayout.unwrapped,
-            viewClass: viewClass ?? View.self,
-            config: configure)
+            sublayout: self.sublayout.unwrapped,
+            viewClass: self.viewClass,
+            config: self.configure)
         super.init(layout: layout)
     }
 
--- a/Sources/ObjCSupport/LOKLabelLayout.swift
+++ b/Sources/ObjCSupport/LOKLabelLayout.swift
@@ -9,6 +9,15 @@
 import UIKit
 
 @objc open class LOKLabelLayout: LOKBaseLayout {
+    @objc public let attributedString: NSAttributedString?
+    @objc public let string: String?
+    @objc public let lineHeight: CGFloat
+    @objc public let font: UIFont
+    @objc public let numberOfLines: Int
+    @objc public let alignment: LOKAlignment
+    @objc public let viewClass: UILabel.Type
+    @objc public let configure: ((UILabel) -> Void)?
+
     @objc public init(attributedString: NSAttributedString,
                       font: UIFont?,
                       lineHeight: CGFloat,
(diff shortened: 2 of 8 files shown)


Sources/ObjCSupport/LOKLabelLayout.swift
```diff
@@ -9,6 +9,9 @@
 import UIKit
 
 @objc open class LOKLabelLayout: LOKBaseLayout {
+    @objc public let attributedString: NSAttributedString?
+    @objc public let text: String?
```
staguer: Should be consistent in the naming of these two properties. Either both end in `text` or both end in `string`.

Sources/ObjCSupport/LOKStackLayout.swift
```diff
@@ -41,6 +41,8 @@ extension LOKStackLayoutDistribution {
 }
 
 @objc open class LOKStackLayout: LOKBaseLayout {
+    @objc public let sublayouts: [LOKLayout]?
```
staguer: Should we expose the other attributes of these layouts as well? Axis and spacing for example. Or is there something special about these properties that makes them useful and necessary?

Sources/ObjCSupport/LOKLabelLayout.swift
nbadakh: Renamed variables.

Sources/ObjCSupport/LOKLabelLayout.swift
```diff
@@ -18,6 +21,8 @@ import UIKit
                       viewReuseId: String?,
                       viewClass: UILabel.Type?,
                       configure: ((UILabel) -> Void)?) {
+        self.attributedText = attributedString
```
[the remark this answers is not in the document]
nbadakh: Fixed

Sources/ObjCSupport/LOKStackLayout.swift
nbadakh: There are computed vars for`flexibility` and `viewReuseId` in `LOKBaseLayout`, so added properties for all remaining layout parameters.

Sources/ObjCSupport/LOKInsetLayout.swift
```diff
@@ -9,12 +9,23 @@
...
     @objc public init(insets: EdgeInsets,
                       alignment: LOKAlignment? = nil,
                       viewReuseId: String? = nil,
                       viewClass: View.Type? = nil,
                       sublayout: LOKLayout,
                       configure: ((View) -> Void)? = nil) {
+        self.insets = insets
+        self.sublayout = sublayout
+        self.alignment = alignment
+        self.viewClass = viewClass
```
staguer: Since this implementation provides defaults for some parameters `alignment` and `viewClass`, it would be good to have the properties reflect that.

Since that would turn some optional properties into non-optional, that'd be a breaking change so it'd be better to get it done with this change, so we don't have to make a breaking change later.

nbadakh: Thanks for pointing this out. Done the required changes.

Sources/ObjCSupport/LOKStackLayout.swift
```diff
@@ -50,15 +58,22 @@ extension LOKStackLayoutDistribution {
                       viewReuseId: String? = nil,
                       sublayouts: [LOKLayout]?,
                       configure: ((View) -> Void)? = nil) {
+        self.axis = axis
+        self.spacing = spacing
+        self.distribution = distribution.distribution != nil ? distribution : .fillFlexing
+        self.sublayouts = sublayouts ?? []
+        self.alignment = alignment ?? .fill
+        self.viewClass = viewClass
```
staguer: Let's make viewClass non-optional here an default to View.self, just like SizeLayout does.

Sources/ObjCSupport/LOKLabelLayout.swift
```diff
@@ -9,6 +9,15 @@
 import UIKit
 
 @objc open class LOKLabelLayout: LOKBaseLayout {
+    @objc public let attributedString: NSAttributedString?
+    @objc public let string: String?
+    @objc public let font: UIFont
+    @objc public let numberOfLines: Int
+    @objc public let alignment: LOKAlignment
```
staguer: Looks like flexibility is missing here.

Also, please add similar properties on LOKSizeLayout, LOKTextViewLayout, LOKOverlayLayout, and LOKButtonLayout. I realize that you don't currently need them yourself, but it doesn't make sense for the library to have them only on some classes and not on other similar classes.

nbadakh: \- `flexibility` and `viewReuseId` (computed) properties are present under `LOKBaseLayout`, hence we haven't added it in child classes.
\- As per your suggestion, added similar properties in `LOKSizeLayout`, `LOKTextViewLayout`, `LOKOverlayLayout` and `LOKButtonLayout` classes.

Sources/ObjCSupport/LOKSizeLayout.swift
```diff
@@ -9,6 +9,15 @@
 import CoreGraphics
 
 @objc open class LOKSizeLayout: LOKBaseLayout {
+    public let maxWidth: CGFloat?
+    public let maxHeight: CGFloat?
+    @objc public let minWidth: CGFloat
+    @objc public let minHeight: CGFloat
+    @objc public let alignment: LOKAlignment?
```
staguer: We should be able to make this alignment property non-optional by constructing a LOKAlignment object from the Swift SizeLayout's alignment property, because in its initializer it computes a default alignment value. This way, after the object is constructed, we can actually reflect the actual non-null alignment value that the layout has.

Sources/ObjCSupport/LOKLabelLayout.swift
staguer: Good point regarding `flexibility` and `viewReuseId`, thanks.

Sources/ObjCSupport/LOKLabelLayout.swift
```diff
@@ -9,6 +9,15 @@
...
+    @objc public let viewClass: UILabel.Type
+    @objc public let configure: ((UILabel) -> Void)?
+    public let lineHeight: CGFloat?
```
staguer: This property, along with `maxWidth` and `maxHeight`, should probably be `@objc` and non-optional.

PR #222, opened 2018-07-24
Added AutoRotateStackLayout

nbadakh:
\- The `AutoRotateStackLayout` displays its contents along the horizontal axis by default.
\- If the contents do not fit along the horizontal axis, the `AutoRotateStackLayout` switchs
   to displaying them along the vertical axis.
\- The `AutoRotateStackLayout` uses to StackLayouts to implement its behavior.
\- Also added `AutoRotateStackViewController` to demonstrate the use of `AutoRotateStackLayout`.
TODO:
Add tests.


nbadakh:
\- Attached screenshots:

landscape_horizontal_axis

portrait_vertical_axis

FabianTerhorst:
This would be the first none essential layout that would go into the library. Wouldn't it make more sense to split it in an sub project?

nbadakh:
Added unit tests

staguer:
@FabianTerhorst you bring up a good point, thank you, we'll consider it.

staguer:
@FabianTerhorst We're considering generalizing this component a bit. It would take an array of layout options and at measurement time it would pick the first one that fits. Sounds like a useful primitive for apps that need to support Dynamic Type and languages with long words. What do you think?

FabianTerhorst:
Sounds good.

PR #229, opened 2018-08-14
HUED-8605 Test & Fix Potential truncation issue in LabelLayout

PrkumariLI:
- Added line break mode support in LabelLayout to calculate correct size based on line break mode.


PrkumariLI:
- Added LabelLayout unit test to verify that size calculation logic considers line break mode while calculating LabelLayout size.

Sources/ObjCSupport/LOKLabelLayout.swift
```diff
@@ -13,11 +13,13 @@ import UIKit
     @objc public let string: String?
     @objc public let lineHeight: CGFloat
     @objc public let font: UIFont
+    @objc public let lineBreakMode: NSLineBreakMode?
     @objc public let numberOfLines: Int
     @objc public let alignment: LOKAlignment
     @objc public let viewClass: UILabel.Type
     @objc public let configure: ((UILabel) -> Void)?
 
+    // TODO: Remove this init once all consumers are switch to `init(attributedString:font:lineBreakMode:lineHeight:numberOfLines:alignment:flexibility:viewReuseId:viewClass:configure:)`
```
staguer: I think we should just add the new parameter, since it's optional.

PrkumariLI: - We have added a new parameter in LOKLabelLayout.swift but it will be a breaking change because every Objective-C caller will have to pass the new parameter else it won't build. Also we made it non-optional parameter as optional enum won't be accessible in Objective-C. 
Let us know if this should be fine.

PrkumariLI:
- Fixed review comments.

p-jadhav:
These changes are being tracked in pull request: https://github.com/linkedin/LayoutKit/pull/260

PR #230, opened 2018-08-16
Fix TextView playground not running in Swift 4.X because of key

tedgonzalez:
Fix TextView playground not running in Swift 4.X because of key


staguer:
Thanks for the fix!

PR #239, opened 2018-10-11
Fix two flexibility issues

hqin-work:
1. LOKOverlayLayoutBuilder's "flexibility" block property doesn't have any effect. To fix it, we pass privateFlexibility to LOKOverlayLayout constructor.
2. LOKInsetLayoutBuilder's "flexibility" block property doesn't have any effect because InsetLayout always takes its sublayout's flexibility as its own flexibility. To fix it, we have InsetLayout constructor take in a flexibility parameter and only use its sublayout's flexibility when that parameter is nil.

### DIFF
--- a/Sources/ObjCSupport/LOKInsetLayout.swift
+++ b/Sources/ObjCSupport/LOKInsetLayout.swift
@@ -11,7 +11,7 @@ import CoreGraphics
 @objc open class LOKInsetLayout: LOKBaseLayout {
     @objc public let insets: EdgeInsets
     @objc public let alignment: LOKAlignment
-    @objc public let flexibility: LOKFlexibility?
+    @objc public let flexibility: LOKFlexibility
     @objc public let viewClass: View.Type
     @objc public let sublayout: LOKLayout
     @objc public let configure: ((View) -> Void)?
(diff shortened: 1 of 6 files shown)


Sources/ObjCSupport/LOKInsetLayout.swift
```diff
@@ -11,24 +11,28 @@ import CoreGraphics
...
                       alignment: LOKAlignment? = nil,
+                      flexibility: LOKFlexibility? =nil,
                       viewReuseId: String? = nil,
                       viewClass: View.Type? = nil,
                       sublayout: LOKLayout,
                       configure: ((View) -> Void)? = nil) {
         self.insets = insets
         self.sublayout = sublayout
         self.alignment = alignment ?? .fill
+        self.flexibility = flexibility
```
staguer: We should probably make this non-optional and have it reflect the actual flexibility value that would be used.

hqin-work: Made `flexibility` here non-optional. But the parameter is still optional (in a different sense of the word) in InsetLayout's constructor.

Sources/ObjCSupport/LOKOverlayLayout.swift
```diff
@@ -34,6 +35,7 @@ import Foundation
             backgroundLayouts: backgroundLayouts?.map { $0.unwrapped } ?? [],
             overlayLayouts: overlayLayouts?.map { $0.unwrapped } ?? [],
             alignment: self.alignment.alignment,
+            flexibility: flexibility?.flexibility ?? .flexible,
```
dgattey: Making this `flexibility?.flexibility ?? primaryLayouts.first?.flexibility ?? .flexible` should maintain backwards compatibility with the previous functionality where we took the first primary layout's flexibility if unspecified.

PR #243, opened 2018-11-01
Fix a recycling bug that purges indirect subviews

Candyroot:
Only walk first level of subviews that are created by LayoutKit.
If a subview isn't created by LayoutKit, then there is no need to walk
subviews of it, as they must not be created by this layout.

### DIFF
--- a/LayoutKitTests/ViewRecyclerTests.swift
+++ b/LayoutKitTests/ViewRecyclerTests.swift
@@ -91,6 +91,30 @@ class ViewRecyclerTests: XCTestCase {
         XCTAssertNotNil(one.superview)
     }
 
+    func testRecycledViewWithoutPurgingIndirectSubviews() {
+        let root = View()
+        let headerView = View()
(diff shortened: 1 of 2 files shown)


staguer:
Oh, this looks like a good improvement, thanks!

I have a couple of points:
1. This avoids traversing rootView -> nonLayoutKitView -> otherLayoutKitViews, right? What if we had rootView -> LayoutKitView -> nonLayoutKitView -> otherLayoutKitViews? Perhaps we should do this check at every level?
2. Do we need to check `$0.isLayoutKitView || $0.viewReuseId != nil` or is it sufficient to just check `$0.isLayoutKitView`?
3. I liked your comment at the start of this Pull Request here, can you please add the comment to the code as well, explaining why it’s important to have the check and what problem it avoids?

Thank you!

Candyroot:
Hi Sergei, Thanks for your review. Your points sound good. I updated my PR to include those changes. Thanks!

PR #244, opened 2018-11-01
Swift 4.2 migration

staguer:
Note: `pod lib lint` requires CocoaPods 1.6 for tvOS pod sanity test to pass


Nadohs:
Are the tests passing locally? Travis is showing errors, but it seems more like its probably just not configured correctly for the 4.2 update.

Nadohs:
Testing on my fork, this fixed the travis issues for me, except for the last step, which was pulling the latest version from cocoapods, which does pulled version of code prior to your syntax updates in this RB... and so failed afte 10 minutes because of that.

ashihaby:
👍

vasilenkoigor:
Hi! Any progress?

PR #245, opened 2018-12-07
Allow embedding layouts in a view hierarchy created from another layout

crleona:
I ended up getting two approaches for this working:

The first marks whichever view is hosting the layout as a layout kit root, and blocks the view recycler from examining the subviews of any view marked similarly.  The code for this is much simpler, but there's no way to clean up an `isLayoutKitRoot` property from the root view, and this would have no effect on view hierarchies created from `makeViews` with a nil view.  (Theoretically, a view hierarchy created without a parent then added to a parent view should still allow view recycling if `makeViews` is called again with the new parent).

The second, implemented here, would be to mark the top level views a layout generates as roots.  This should resolve the issues the first approach had at the cost of some additional complexity.  It makes the assumption that all direct subviews marked as root of a parent view passed in to `makeViews` belong to that layout or should be removed, but that should be consistent with the existing behavior of ViewRecycler.  I'm at a loss as to what the automatically created container view for a parentless layout with multiple views should be marked as, but I think leaving it as is will match the current behavior.

There's also a small bug fix for associated objects - namely, they should be stored as copy, nonatomic.


Sources/ViewRecycler.swift
```diff
@@ -96,23 +107,36 @@ class ViewRecycler {
...
+        case managed
+        // Indicates the view is managed by LayoutKit but was generated by another layout and should not be modified.
+        case root
+    }
+
     /// Calls visitor for each transitive subview.
-    func walkSubviews(visitor: (View) -> Void) {
-        for subview in subviews {
+    func walkNonRootSubviews(visitor: (View) -> Void) {
+        for subview in subviews where subview.type != .root {
```
Candyroot: Is it necessary to walk unmanaged subviews? Won't the subviews of the subview be either unmanaged or root ones?

staguer: This is a good point. Let's tackle it as a separate change.

PR #246, opened 2019-01-19
update to swift 4.2

HebaAtallah:


FabianTerhorst:
There is already an pr for this #244

PR #257, opened 2019-05-15
Migrated swift version to 5.0

s-vartak:
\- Updated the swift version to 5.0
-    Added changes for accessing enum
-   Converted enums in ObjC to [frozen enums](https://useyourloaf.com/blog/swift-5-frozen-enums/)
-   Added case `@unknown default` when using system's non-frozen enum(ObjC) in switch-case

\- Made travis use xcode10.2

\- Updated `build.sh` file to use updated SDKs and simulators
- `iphonesimulator11.3` -> `iphonesimulator12.2`
- `macosx10.13` -> `macosx10.14`
- `appletvsimulator11.3` -> `appletvsimulator12.2`
- `Apple TV 1080p` -> `Apple TV 4K (at 1080p)`

\- Added `-disable-concurrent-destination-testing` flag to xcodebuild commands to disable concurrency issues happening on XCODE 10.2 target.


LayoutKit.xcodeproj/project.pbxproj
```diff
@@ -1908,6 +1914,7 @@
 				SKIP_INSTALL = YES;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 5.0;
```
crleona: if swift version is set at the project/workspace level, it should be inherited into all of these configurations - these additions can probably be deleted.

s-vartak: Yes, you are right. Removed these unnecessary changes.

PR #260, opened 2019-06-07
Fixed Potential truncation issue in LabelLayout

p-jadhav:
- Added line break mode support in LabelLayout to calculate correct size based on line break mode.
- Added LabelLayout unit test to verify that size calculation logic considers line break mode while calculating LabelLayout size.
- These changes are duplicate of pull request https://github.com/linkedin/LayoutKit/pull/229


p-jadhav:
Truncation_Issue
Truncation_Issue_Fixed

